PR #10, opened 2020-01-29
Finishing touches for publishing to maven central

ogolberg:


buildSrc/src/main/kotlin/RemotePublishing.kt
```diff
@@ -106,12 +109,17 @@ fun Project.enablePublishing(defaultJars: Boolean = true) {
             archiveClassifier.set("sources")
         }
 
+        tasks.register<Jar>("javadocJar") {
+            from("$rootDir/README.md")
```
aprotokt-toast: Does this include the README? I'm not sure I've ever seen this before.

ogolberg: see https://central.sonatype.org/pages/requirements.html - javadoc jar is required, and they recommend packaging up the readme into the javadocs jar when javadocs are not available (dokka didn't work immediately - we can look into it later)

PR #14, opened 2020-02-13
A quick and dirty workaround for Windows

ogolberg:
On Windows, `protoc` expects a `/`-separated path to the plugin binary. The binary needs to include the extension (`.bat` in our case).

Tested on Windows 10.

### DIFF
--- a/buildSrc/src/main/kotlin/com/toasttab/protokt/shared/ProtobufBuild.kt
+++ b/buildSrc/src/main/kotlin/com/toasttab/protokt/shared/ProtobufBuild.kt
@@ -24,6 +24,7 @@ import com.google.protobuf.gradle.protoc
 import com.google.protobuf.gradle.remove
 import org.gradle.api.Project
 import org.gradle.api.plugins.JavaPluginConvention
+import org.gradle.internal.os.OperatingSystem
 import org.gradle.kotlin.dsl.apply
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/kotlin/com/toasttab/protokt/shared/ProtobufBuild.kt
```diff
@@ -65,3 +66,10 @@ internal fun configureProtobufPlugin(project: Project, ext: ProtoktExtension, bi
         }
     }
 }
+
+private fun normalizePath(binaryPath: String) =
+    if (OperatingSystem.current().isWindows) {
+        binaryPath.replace('\\', '/') + ".bat"
```
aprotokt-toast: I'd prefer we just built the path with File.separator here: https://github.com/open-toast/protokt/blob/master/buildSrc/src/main/kotlin/LocalProtoktBuild.kt#L23.

ogolberg: yes, and then you still have to replace \ with / because protoc wants /  :rage3:

aprotokt-toast: Can we leave a comment to that effect?

PR #18, opened 2020-02-15
Fix wrappers for all primitive types

aprotokt-toast:
When introducing the `wrapped` field on Converters for protokt to look up, byte arrays were inadvertently the only primitive type that worked, since `Class.forName(kotlin.X)` doesn't actually work. Those types have to be represented by their Java class names.

Related issue: https://youtrack.jetbrains.com/issue/KT-10440

### DIFF
--- a/protokt-codegen/src/main/resources/renderers.stg
+++ b/protokt-codegen/src/main/resources/renderers.stg
@@ -129,6 +129,23 @@ isScalar ::= [
     default: false
 ]
 
+javaTypeNames ::= [
+    "BOOL": "Boolean",
+    "DOUBLE": "Double",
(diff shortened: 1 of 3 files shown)


protokt-codegen/src/main/resources/renderers.stg
```diff
@@ -129,6 +129,23 @@ isScalar ::= [
     default: false
 ]
 
+javaTypeNames ::= [
```
ogolberg: I thought this was moving to the options module

aprotokt-toast: It's headed there (in #19).

PR #19, opened 2020-02-18
Represent well known types with Kotlin nullable types

aprotokt-toast:
Use wrapper types to represent all well-known types with their Kotlin nullable counterparts.

`StringValue` becomes `String?`, etc.

In order to wrap the well-known types by default, the modules that used to be called `protokt-core` and `protokt-extensions-simple` (just for wkt wrappers) have to always be available to the compiler's class path. And `protokt-core`'s generated well-known types have to be available to `protokt-extensions-simple`.

Module reorganization was necessary:
- the well-known types are now generated in a module called `protokt-wkt`.
- `protokt-core`, which is the module every external consumer will use, api-depends on `protokt-runtime`, `protokt-exensions-api`, and `protokt-wkt`. It contains converters for each well-known type.


buildSrc/src/main/kotlin/com/toasttab/protokt/shared/ProtoktExtension.kt
```diff
@@ -17,7 +17,7 @@ package com.toasttab.protokt.shared
 
 import java.util.jar.JarInputStream
 
-const val DEFAULT_PROTOBUF_VERSION = "3.11.0"
+const val DEFAULT_PROTOBUF_VERSION = "3.11.3"
```
ogolberg: 3.11.4  😄

buildSrc/src/main/kotlin/com/toasttab/protokt/shared/ProtoktBuild.kt
```diff
@@ -34,3 +37,22 @@ private fun createExtensionConfiguration(project: Project) {
         extendsFrom(extensionsConfiguration)
     }
 }
+
+private fun addProtoktCoreDependency(project: Project, ext: ProtoktExtension) {
+    if (project.name !in setOf("protokt-wkt", "protokt-core")) {
```
ogolberg: "shared" build code should not have any internal protokt build logic; this state should be pulled into protokt-specific part of buildSrc and passed into configureProtokt as a variable

aprotokt-toast: Done.

buildSrc/src/main/kotlin/com/toasttab/protokt/shared/ProtoktExtension.kt
aprotokt-toast: Done.

PR #28, opened 2020-04-03
Allow usage of java_package option (and respect it by default)

andrewparmet:
This is a prerequisite for service generation of existing protos (e.g. health check) assuming we want to leverage grpc-kotlin, which seems reasonable - grpc-kotlin looks for descriptors in the Java package when specified. If we want grpc-kotlin to respect the kotlin_package option we'd have to do additional work (or ask them to support our option), but we don't have to worry about that for now.

It's also nice to have this option in general for external dependencies to have normal-looking packages, for example the third-party proto-google-common-protos.


buildSrc/src/main/kotlin/LocalProtoktBuild.kt
```diff
@@ -21,6 +21,9 @@ import org.gradle.kotlin.dsl.dependencies
 import org.gradle.kotlin.dsl.project
 import org.gradle.kotlin.dsl.withType
 
+val Project.buildSrcClasses
```
ogolberg, line 24: val -> fun

buildSrc/src/main/kotlin/com/toasttab/protokt/shared/ProtoktExtension.kt
```diff
@@ -23,7 +23,8 @@ const val MANIFEST_VERSION_PROPERTY = "Implementation-Version"
 open class ProtoktExtension {
     var protocVersion = DEFAULT_PROTOBUF_VERSION
     var toolsVersion: String? = null
-    var publishProto: Boolean = false
+    var publishProto = false
+    var respectJavaPackage = true
```
ogolberg, line 27: any reason to have this configurable vs always fall back to java_package? (like scalapb, for example)

buildSrc/src/main/kotlin/LocalProtoktBuild.kt
andrewparmet: I don't mind making this change. Why do you feel strongly about this?

buildSrc/src/main/kotlin/com/toasttab/protokt/shared/ProtoktExtension.kt
andrewparmet: My thoughts were migration - if a project is using protobuf-java and wants to transition to protokt, always falling back on java_package means any code used from libraries (to be replaced with a protokt-generated library) can't coexist. Once migrated, they can re-enable the option to get the proper package names back.

ogolberg: sounds good

buildSrc/src/main/kotlin/LocalProtoktBuild.kt
ogolberg: internally, we prefer funs; in this codebase we're not consistent about it

since the value effectively never changes, I am ok leaving it as a val

PR #29, opened 2020-04-04
Generate gRPC Method Descriptors

andrewparmet:
Generates service and method descriptors:

```kotlin
object OperationsGrpc {
    const val SERVICE_NAME = "google.longrunning.Operations"

    val serviceDescriptor: ServiceDescriptor by lazy {
        ServiceDescriptor.newBuilder(SERVICE_NAME)
            .addMethod(listOperationsMethod)
            .build()
    }

    val listOperationsMethod: MethodDescriptor<ListOperationsRequest, ListOperationsResponse> by lazy {
        MethodDescriptor.newBuilder<ListOperationsRequest, ListOperationsResponse>()
            .setType(MethodType.UNARY)
            .setFullMethodName(generateFullMethodName(SERVICE_NAME, "ListOperations"))
            .setRequestMarshaller(KtMarshaller(ListOperationsRequest))
            .setResponseMarshaller(KtMarshaller(ListOperationsResponse))
            .build()
    }
}
```


protokt-codegen/src/main/resources/services.stg
```diff
@@ -13,30 +13,60 @@
...
 
->>
+    <methods : { m | val <m.lowerName>Method: MethodDescriptor\<<m.in>, <m.out>\> by lazy {
+    <m.body><\n><\\>
+    \}}; separator="\n\n">
 
-descriptorSupplier(fqcn) ::= <<
+    @JvmName("descriptor")
+    @JvmSynthetic
+    fun getServiceDescriptor() =
```
ogolberg, line 32: would `@get:JvmName` work?

andrewparmet: `This annotation is not applicable to target 'member function' and use site target '@get'`

andrewparmet: Strictly speaking the `@JvmSynthetic` annotation isn't need for this to compile, it's just to hide it so the getter is the only visible item from Java.

ogolberg: I meant `@get:JvmName`  on the `val`

andrewparmet: Ah - the catch is grpc-kotlin calls the `get()` functions _from Kotlin_, so it has to be present as a function in Kotlin.

andrewparmet: https://github.com/grpc/grpc-kotlin/blob/master/compiler/src/test/java/io/grpc/kotlin/generator/HelloWorldProtoGrpcKt.expected#L57

andrewparmet: So yes - `@get:JvmName` on the `val`s works, but then you have an extra visible member from Java. I don't know whether we care.

PR #35, opened 2020-04-27
Gradle 6.3, handle errors from effects, remove templates.kt

andrewparmet:
`templates.kt` was a mess.

Also improves "type safety" with StringTemplates.

There's still a little more simplification I'd like to do, but that's more of an actual behavior change as opposed to this, which while large is mostly just cosmetic.


protokt-codegen/src/main/kotlin/com/toasttab/protokt/codegen/template/TemplateVariable.kt
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.toasttab.protokt.codegen.template
+
+interface TemplateVariable {
```
andrewparmet, line 18: I think I can get rid of this completely, next round.

protokt-codegen/src/main/resources/renderers.stg
```diff
@@ -141,7 +141,7 @@ isGeneratedType ::= ["MESSAGE": true, "ENUM": true, default: false]
 
 isBytes ::= ["BYTES": true, "STRING": true, default: false]
 
-type2nativeF(type) ::= "<kotlinTypeNames.(type)>"
+typeTonativeF(type) ::= "<kotlinTypeNames.(type)>"
```
ogolberg, line 144: `typeToNativeF` ?

protokt-codegen/src/main/kotlin/com/toasttab/protokt/codegen/template/UtilityParams.kt
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright (c) 2019 Toast Inc.
```
ogolberg, line 2: 2020

PR #37, opened 2020-04-29
Implement custom marshallers; fix import bug

andrewparmet:
- Allow custom marshallers for request and response types
- Fix bug resolving imports for classes that also exist in the package declared by a proto file (don't bother importing)
- Unify testing package strategy


protokt-codegen/src/main/kotlin/com/toasttab/protokt/codegen/impl/ImportFilterer.kt
```diff
@@ -15,16 +15,29 @@
...
+) {
     private val internalPackages = setOf(PPackage.PROTOKT, PPackage.PROTOKT_RT)
 
     fun filterDuplicateSimpleNames(imports: Sequence<Import>): Set<Import> {
         val classImports = imports.filterIsInstance<Import.Class>().toSet()
         val nonDuplicateImports = mutableSetOf<Import.Class>()
 
-        classImports.forEach { import ->
+        for (import in classImports) {
+            if (getClass(PClass(import.pClass.simpleName, pkg, None)).isDefined()) {
```
ogolberg, line 35: filter?

andrewparmet: oh fair enough

PR #38, opened 2020-04-30
Use nativeTypeName in deserializer expression for oneofs

andrewparmet:
Additionally, clean up StandardField `name`.

cc @jsanford-toast

Given:
```protobuf
message MessageName {
  oneof message_name {
    string child = 1;
  }
}
```

We were getting:
```kotlin
when (deserializer.readTag()) {
    ...
    10 -> messageName = 
        MessageName.Child(deserializer.readString())
    ...
}
```
and we want:
```kotlin
when (deserializer.readTag()) {
    ...
    10 -> messageName = 
        MessageName_.Child(deserializer.readString())
    ...
}
```

Alternatively we can ditch the underscores. The following would need fully qualified names:
```kotlin
override fun equals(other: Any?): Boolean =
    other is com.toasttab.protokt.testing.rt.oneof.MessageName &&
```

and

```kotlin
companion object Deserializer : KtDeserializer<com.toasttab.protokt.testing.rt.oneof.MessageName>, (MessageNameDsl.() -> Unit) -> com.toasttab.protokt.testing.rt.oneof.MessageName {
    override fun deserialize(deserializer: KtMessageDeserializer): com.toasttab.protokt.testing.rt.oneof.MessageName {
```

... but there may be further complications down the line if more messages are nested, especially if they have the same name. The amount of full qualification necessary may just increase forever. It's cleaner generated code to use underscores, and cleaner calling code to fully qualify where needed.

### DIFF
--- a/protokt-codegen/src/main/kotlin/com/toasttab/protokt/codegen/impl/AnnotatorUtils.kt
+++ b/protokt-codegen/src/main/kotlin/com/toasttab/protokt/codegen/impl/AnnotatorUtils.kt
@@ -41,7 +41,7 @@ internal fun oneOfScope(f: Oneof, type: String, ctx: Context) =
         ctx.stripRootMessageNamePrefix(
             ConcatWithScope.render(
                 scope = type,
-                value = f.nativeTypeName
+                value = f.name
             )
         )
     )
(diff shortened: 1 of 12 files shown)


ogolberg:
would it work with this though?

```
message MessageName_ {
  oneof message_name {
    string child = 1;
  }
}
```

scalapb nests and fully qualifies like so

```
final case class MessageName(
    messageName: com.toasttab.protokt.testing.rt.oneof.shared_name.MessageName.MessageName = ...
```

andrewparmet:
It works fine with that one:
```kotlin
class MessageName_
private constructor(
    val messageName: MessageName__?,
    val unknown: Map<Int, Unknown> = emptyMap()
) : KtMessage {
    sealed class MessageName__ {
        data class Child(
            val child: String
        ) : MessageName__()
    }
```

This is just fixing the fact that the nesting strategy wasn't being followed - the name with underscores is already being generated that way.

We can fully qualify instead, but I think I'd prefer to do that as a follow up to this.

PR #59, opened 2020-06-03
tag generated files with the version of protokt that generated them

andrewparmet:


buildSrc/src/main/kotlin/com/toasttab/protokt/shared/ProtoktExtension.kt
```diff
@@ -16,28 +16,30 @@
...
-                        .manifest
-                        .mainAttributes
-                        .getValue(MANIFEST_VERSION_PROPERTY)
-                }
+        getProtoktVersion(ProtoktExtension::class)
     }
 }
+
+fun getProtoktVersion(klass: KClass<*>): String =
+    klass.java
```
ogolberg, line 35: this code path is only used by the gradle plugin, not the local protokt build

so it can be moved into a tiny shared module instead of packaging buildSrc classes into codegen

andrewparmet: We still need access to the constants for options. That happened to work without adding to the JAR task, so those can stay put?

ogolberg: also you can forego manifest generation and parsing and add something like https://github.com/crotwell/version-class to the codegen and plugin projects

andrewparmet: That plugin didn't work OOB and my best finagling with SourceSets couldn't get it working, so I'm going to defer that for now. I've extracted to a small module.

andrewparmet: ... which unfortunately, must be published.

PR #64, opened 2020-06-05
Cleanup unknown field representation

### LINKED ISSUE #34: Consider reworking unknown field representation
Consider: 
- not being a map, so we don't have to wrap or make copies of the unknown fields object
- ~not storing everything as a ListVal? It looks like everything, even non-repeated fields, is stored in a repeated-compatible way. I'm not sure if that in the spec.~ it's in the spec

andrewparmet:
Looking through the implementation of unknown fields, it's entirely non-obvious that it's correct. `Unknown` was overloaded - it both represents a list of unknown values as retained by the completed set of unknown fields and each field in that set. Instead, model the implementation more closely on the protobuf-java UnknownFieldSet.

Revealed a bug in equality of unknown fields based on strings or byte arrays.

Question to consider: should we bother trying to keep this backwards compatible? I think it might be possible to leave and deprecate old functions, since Unknown didn't actually change too much. ListVal would have to stick around as well.

Fixes #34.

### DIFF
--- a/protokt-codegen/src/main/resources/message.stg
+++ b/protokt-codegen/src/main/resources/message.stg
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 
-import "companion.stg"
+import "deserializer.stg"
 import "dsl.stg"
 import "options.stg"
 import "renderers.stg"
@@ -124,7 +124,7 @@ private constructor(
 
     <dsl()>
 
-    <companion()>
+    <deserializer()>
     <if (nested)>
 
     <nested:{n | <n>}; separator="\n"><\\>
--- a/protokt-codegen/src/main/resources/message.stg
+++ b/protokt-codegen/src/main/resources/message.stg
@@ -31,14 +31,14 @@ private constructor(
     <messagePropertyOverrides()><\\>
     val <p.name>: <p.propertyType><\\>
     }; separator=",\n"><if (properties)>,<endif>
-    val unknown: Map\<Int, Unknown> = emptyMap()
+    val unknownFields: UnknownFieldSet = UnknownFieldSet.empty()
 ) : KtMessage<messageImplements()> {
     <if (oneofs)>
     <oneofs:{it |<it>}; separator="\n\n">
 
     <else>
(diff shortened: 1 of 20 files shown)


protokt-codegen/src/main/resources/message.stg
```diff
@@ -31,13 +31,13 @@ private constructor(
     <messagePropertyOverrides()><\\>
     val <p.name>: <p.propertyType><\\>
     }; separator=",\n"><if (properties)>,<endif>
-    val unknown: Map\<Int, Unknown> = emptyMap()
+    val unknownFields: UnknownFieldSet = UnknownFieldSet.empty()
```
andrewparmet, line 34: This default value is unneeded in Kotlin, but it is needed for our hobbled support of JSON deserialization.

protokt-runtime/src/main/kotlin/com/toasttab/protokt/rt/UnknownFieldSet.kt
```diff
@@ -0,0 +1,157 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.toasttab.protokt.rt
+
+class UnknownFieldSet
+private constructor(
+    val unknownFields: Map<Int, Field>
```
ogolberg, line 20: can this be inline so that we don't create an extra wrapper object?

andrewparmet: We'd have to expose the constructor. Most of the time the unknown field set will be the singleton empty object - API-wise, how do we feel about saving the extra object and exposing the constructor for this "micro-optimization" (?) vs. not exposing it?

andrewparmet: Per discussion - we can revisit later.

PR #71, opened 2020-06-14
add testProtoktExtensions

andrewparmet:
Allows protobuf in src/test/proto to use extensions not available to main code.

### DIFF
--- a/buildSrc/src/main/kotlin/com/toasttab/protokt/gradle/ProtobufBuild.kt
+++ b/buildSrc/src/main/kotlin/com/toasttab/protokt/gradle/ProtobufBuild.kt
@@ -15,6 +15,7 @@
 
 package com.toasttab.protokt.gradle
 
+import com.google.protobuf.gradle.GenerateProtoTask
 import com.google.protobuf.gradle.builtins
 import com.google.protobuf.gradle.generateProtoTasks
 import com.google.protobuf.gradle.id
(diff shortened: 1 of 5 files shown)


buildSrc/src/main/kotlin/com/toasttab/protokt/gradle/ProtobufBuild.kt
```diff
@@ -81,6 +75,11 @@ internal fun configureProtobufPlugin(project: Project, ext: ProtoktExtension, bi
     }
 }
 
+private fun Project.extraClasspath() =
+    (configurations.getByName(EXTENSIONS) + configurations.getByName(TEST_EXTENSIONS))
```
andrewparmet, line 79: I'm not sure exactly how this works, but I've verified that dependencies declared only in `testProtoktExtensions` are not available during `generateProto`.

ogolberg: I would expect testProtoktExtensions to always be added to KOTLIN_EXTRA_CLASSPATH

andrewparmet: It is always added - simple collection merging here. What I'm saying is something declared in testProtoktExtensions shouldn't be available for protobuf in main generation, like if you want to use an extension in main protobuf but declare it in testProtoktExtensions.

ogolberg: right, it shouldn't

but it seems like this change will expose testExtensions dependencies to generateProto; what am I missing?

andrewparmet: I was wrong - just like it looked, those extensions were available for main code generation. If there was a reference to classes, then compilation would fail, as the `api` configuration did not pick up the dependencies. I've reworked it to only include the extensions when appropriate: https://github.com/open-toast/protokt/pull/71/commits/ae7880253915de34575cce1ffe04e1038515c2b0

PR #74, opened 2020-06-21
add imports for map key and value wrappers

andrewparmet:


protokt-codegen/src/main/kotlin/com/toasttab/protokt/codegen/impl/StandardFieldImportResolver.kt
```diff
@@ -81,18 +84,31 @@ class StandardFieldImportResolver(
...
-                set.add(pclass(wrapper))
-                set.add(pclass(converter(wrapper, wrapped, ctx)::class))
-            }
-        )
+            )
+        }
+
+        addWrappers { p, c, e, s -> f.foldFieldWrap(p, c, e, s) }
+        addWrappers { p, c, e, s -> f.foldKeyWrap(p, c, e, s) }
+        addWrappers { p, c, e, s -> f.foldValueWrap(p, c, e, s) }
```
ogolberg, line 111: 🤦

andrewparmet: I think I found a better way, bear with me...

PR #76, opened 2020-06-22
Generate error messages at throw-time; consistent behavior for non-null fields in DSL and deserialization

### LINKED ISSUE #32: Consider not creating custom strings for errors
Consider calling methods to generate those strings when they're needed.

https://jakewharton.com/the-economics-of-generated-code/

andrewparmet:
Extracts error message generation to be computed when thrown rather than a literal.

Also makes creation consistent between the DSL and deserialization. Right now, if a wrapped scalar (i.e., bytes/UUID) is not present, we attempt to deserialize from the default in the constructor. But if it's not present in the DSL, we throw. I think the two should have consistent behavior and that allowing construction from a missing property is correct, since e.g. java.util.Date can be constructed from the default without issue (and in fact would be serialized that way).

```kotlin
// constructor in deserializer today
uuid ?: UuidConverter.wrap(Bytes.empty.bytes)

// constructor in DSL today
requireNotNull(uuid) {
    "uuid is a wrapped primitive, was not specified, and has no default value"
}
```

Fixes #32.


protokt-codegen/src/main/resources/renderers.stg
```diff
@@ -118,15 +118,18 @@ deserializeValue(p) ::= <%
     <endif>
 %>
 
-requireNotNullMessage(p) ::= <%
-    "<p.name> specified nonnull with (protokt.<if (p.oneOf)>oneof<else>property<endif>).non_null but was null"
-%>
+requireNotNullMessage(p) ::= <<
+requireNotNull(<p.name>) {
+    StringBuilder("<p.name>")
```
ogolberg, line 123: why do we need a stringbuilder?

andrewparmet: If we just use concatenation, the compiler will "optimize" and do the concatenation at compile-time, which we don't want. This whole PR is an optimization geared towards Android - attempting to avoid creation of rarely needed long string literals in generated code, which tends to proliferate.

andrewparmet: In an earlier commit this is a function exposed by the runtime, but I thought it's probably better not to expose a method just for this. Maybe this can be pulled into a locally generated method?

andrewparmet: The Kotlin compiler generates another string for its intrinsics:
```java
var26 = (new StringBuilder("socketAddress")).append(" specified nonnull with (protokt.property).non_null but was null");
Intrinsics.checkExpressionValueIsNotNull(var26, "StringBuilder(\"socketAdd…).non_null but was null\")");
```

But this at least allows the protokt string to be reused.

ogolberg: you could pull the shared parts of errors messages into runtime and generate something like

```
"uuid ${Errors.IS_A_MISSING_WRAPPED_PRIMITIVE}"
```

PR #86, opened 2020-10-02
Strip enum type name prefix when all enum values share it

andrewparmet:


protokt-codegen/src/main/kotlin/com/toasttab/protokt/codegen/protoc/Protocol.kt
```diff
@@ -36,24 +36,13 @@ import com.google.protobuf.DescriptorProtos.FieldDescriptorProto.Label.LABEL_REP
...
-import com.google.protobuf.GeneratedMessage.GeneratedExtension
-import com.google.protobuf.GeneratedMessageV3.ExtendableMessage
 import com.toasttab.protokt.codegen.impl.STAnnotator.rootGoogleProto
 import com.toasttab.protokt.codegen.impl.overrideGoogleProtobuf
 import com.toasttab.protokt.codegen.impl.resolveMapEntry
 import com.toasttab.protokt.codegen.model.FieldType
 import com.toasttab.protokt.codegen.model.PClass
 import com.toasttab.protokt.ext.Protokt
 
-private fun <I, E : ExtendableMessage<E>> E.extensionOrDefault(
```
andrewparmet, line 48: From my inspection, `getExtension` returns the default value when absent, not null.

ogolberg (approved):

PR #104, opened 2021-04-07
Use a more up-to-date kts syntax for plugin/task/dependency configuration

ogolberg:


buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -56,7 +56,6 @@ object libraries {
 
     const val kollection = "com.github.andrewoma.dexx:kollection:${versions.kollection}"
 
-    const val kotlinPlugin = "org.jetbrains.kotlin:kotlin-gradle-plugin:${versions.kotlin}"
     const val kotlinReflect = "org.jetbrains.kotlin:kotlin-reflect:${versions.kotlin}"
     const val kotlinStdlib = "org.jetbrains.kotlin:kotlin-stdlib:${versions.kotlin}"
```
andrewparmet, line 60: We can probably use the `kotlin("foo")` versions of these as well.

ogolberg: I am not happy we even depend on reflect. But I will look into that separately.

PR #105, opened 2021-05-06
Implement reflective access to descriptors

andrewparmet:
Implements encoding of FileDescriptorProtos in their associated Kotlin files and access to them through the companion objects of each generated message type.

I did not port the full protobuf-java reflection API - if you really want, you can serialize the descriptors and deserialize them using standard protobuf-java to use that API. All you need to do that is the file descriptor with its dependencies - which is implemented here. The declaration for that transformation is:

``` java
public static FileDescriptor buildFrom(
  final FileDescriptorProto proto,
  final FileDescriptor[] dependencies,
  final boolean allowUnknownDependencies) throws ...
```

so you could call:

``` kotlin
fun FileDescriptor.toProtobufJavaDescriptor(): Descriptors.FileDescriptor =
    Descriptors.FileDescriptor.buildFrom(
        DescriptorProtos.FileDescriptorProto.parseFrom(proto.serialize()),
        dependencies.map { it.toProtobufJavaDescriptor() }.toTypedArray(),
        true
    )
```

and you should be good to go. I've written a test that exercises this transformation.

I mimic protobuf-java's naming scheme for descriptor objects.

I've written this so that descriptors load lazily. In protobuf-java they're loaded statically. Not sure which we prefer.

`Any` packing and unpacking is still done reflectively. I've left the generated message annotation lookup around for backwards compatibility. Implementing non-reflective lookup will involve some sort of hackery. In protobuf-java they simply expose an instance method on every generated type called `getDescriptor()`. I'm not sure how they work the dependency graph since that `Descriptor` type references a generated type itself. I guess we could depend on an older published version of protokt in the runtime project? Not sure if that would work.

Encoding the descriptors is easier in Kotlin than in Java - Kotlin specifies that files are encoded in UTF-8, so any non-escape character can simply be printed as-is. You'll be at the mercy of your text editor but the file will be valid!

Here's a sample of the newly generated code:

``` kotlin
// any.kt

class Any : KtMessage {
    // ...

    companion object Deserializer : KtDeserializer<Any>, (AnyDsl.() -> Unit) -> Any {
        // ...

        val descriptor by lazy {
            google_protobuf_any.descriptor.messageTypes[0]
        }
    }
}

object AnyProto {
    val descriptor by lazy {
        val descriptorData = arrayOf(
            "\ngoogle/protobuf/any.protogoogle.prot" +
            "obuf\"&\nAny\n\btype_url (\t\r\nvalue" +
            " (Bv\ncom.google.protobufB\bAnyProtoPZ" +
            ",google.golang.org/protobuf/types/known/" +
            "anypbﾢGPBﾪGoogle.Protobuf.WellKnownT" +
            "ypesbproto3"
        )

        FileDescriptor.fromData(
            descriptorData,
            listOf()
        )
    }
}
```


protokt-codegen/src/main/kotlin/com/toasttab/protokt/Main.kt
```diff
@@ -73,6 +80,15 @@ private fun generate(
                     protoFileList,
                     respectJavaPackage(params)
                 ),
+                protoFileList.associate {
+                    it.name to
+                        resolvePackage(
+                            it,
+                            it.name !in filesToGenerate ||
```
andrewparmet, line 87: If we're ignoring the Java package option, we still need to reference any descriptor dependencies _not_ generated in this run that _did_ respect the Java package.

protokt-core/build.gradle.kts
```diff
@@ -34,3 +35,11 @@ dependencies {
 
     add("kapt", libraries.autoService)
 }
+
+sourceSets {
+    main {
+        proto {
+            srcDir("../protokt-runtime/src/main/resources")
```
andrewparmet, line 42: Generate `protokt.proto` for descriptor dependencies.

testing/runtime-tests/build.gradle.kts
```diff
@@ -20,6 +20,7 @@ pureKotlin()
 
 protokt {
     generateGrpc = true
+    respectJavaPackage = false
```
andrewparmet, line 23: Descriptor objects were conflicting with each other.

ogolberg (approved):

PR #106, opened 2021-05-10
update some dependencies

andrewparmet:
Three dependencies are intertwined and must be updated together - Kotlin (to 1.4.x), Arrow (which pulls in Kotlin 1.3.x), and Gradle (which pulls in Kotlin 1.3.x before 6.8 and 1.4.x after).

That'll be a different PR.


beng-toast (approved):
assuming this is a cumulative PR, LGTM

PR #109, opened 2021-05-11
Upgrade Kotlin, Gradle, and Arrow

andrewparmet:
Arrow removed a bunch of things we were using. They can all be replaced with Kotlin built-ins relatively painlessly.


beng-toast (commented):
This looks good. I am wondering if we should just remove arrow dependency altogether. Thoughts ?

andrewparmet:
That's a bit more invasive. I wanted this PR to be the minimal change required to upgrade Kotlin and Gradle, and since Arrow has that explicit dependency on 1.3, it had to be upgraded as well.

ogolberg (approved):

PR #110, opened 2021-05-12
Fix weird Kotlin compiler bug

andrewparmet:
Using a custom getter breaks for some reason I don't understand:

```
java.lang.NoSuchFieldError: size
	at com.toasttab.protokt.rt.UnknownFieldSet$Field.equals(UnknownFieldSet.kt:93)
	at java.base/java.util.AbstractMap.equals(AbstractMap.java:493)
	at java.base/java.util.Collections$UnmodifiableMap.equals(Collections.java:1491)
	at kotlin.jvm.internal.Intrinsics.areEqual(Intrinsics.java:167)
	at com.toasttab.protokt.rt.UnknownFieldSet.equals(UnknownFieldSet.kt:26)
```

but backing it with a real field works.

Also get rid of old conformance files I forgot to clean up.


ogolberg:
should we convert it to a method to avoid storing a field?

andrewparmet:
We could, but then we have to support a method in the runtime vs. a property we can change the implementation of. I think a property is more idiomatic. What about a lambda instead of a field reference?

ogolberg (approved):

PR #111, opened 2021-05-20
Upgrade gradle, fix implicit task dependency, and fix Java targeting

andrewparmet:
The `.module` files were depending on Java 11 by default since that's what we use to run the publish task:

```
{
      "name": "apiElements",
      "attributes": {
        "org.gradle.category": "library",
        "org.gradle.dependency.bundling": "external",
        "org.gradle.jvm.version": 11,
        "org.gradle.libraryelements": "jar",
        "org.gradle.usage": "java-api",
        "org.jetbrains.kotlin.localToProject": "public",
        "org.jetbrains.kotlin.platform.type": "jvm"
      },
```

Also fixes this warning that appears newly in Gradle 7:
```
Execution optimizations have been disabled for task ':protokt-core:sourcesJar' to ensure correctness due to the following reasons:
  - Gradle detected a problem with the following location: '/Users/andrewparmet/protokt/protokt-core/build/generated-sources/main/protokt'. Reason: Task ':protokt-core:sourcesJar' uses this output of task ':protokt-core:generateProto' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. Please refer to https://docs.gradle.org/7.0.1/userguide/validation_problems.html#implicit_dependency for more details about this problem.
```

### DIFF
--- a/buildSrc/src/main/kotlin/RemotePublishing.kt
+++ b/buildSrc/src/main/kotlin/RemotePublishing.kt
@@ -110,6 +110,14 @@ fun Project.enablePublishing(defaultJars: Boolean = true) {
             withSourcesJar()
         }
 
+        afterEvaluate {
+            if (tasks.findByName("generateProto") != null) {
+                tasks.named("sourcesJar").configure {
+                    dependsOn("generateProto")
+                }
+            }
+        }
+
         tasks.register<Jar>("javadocJar") {
             from("$rootDir/README.md")
             archiveClassifier.set("javadoc")
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -72,4 +72,9 @@ subprojects {
             }
         }
     }
+
+    configure<JavaPluginExtension> {
+        sourceCompatibility = JavaVersion.VERSION_1_8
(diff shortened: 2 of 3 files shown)


buildSrc/src/main/kotlin/RemotePublishing.kt
```diff
@@ -110,6 +110,14 @@ fun Project.enablePublishing(defaultJars: Boolean = true) {
             withSourcesJar()
         }
 
+        afterEvaluate {
```
ogolberg, line 113: just highlighting that `afterEvaluate` is considered an anti-pattern and `findByName` is eager (see https://docs.gradle.org/current/userguide/task_configuration_avoidance.html)

but I don't know of a better way to accomplish this

ogolberg (approved):

buildSrc/src/main/kotlin/RemotePublishing.kt
andrewparmet: I think we have to use `afterEvaluate` because of the protobuf-gradle-plugin, unfortunately.

PR #118, opened 2021-07-16
Use sumOf instead of sumBy; upgrade Gradle

### LINKED ISSUE #117: generation of `messageSize()` uses deprecated `sumBy()` function
Apparently, the generation of message types, or more precisely, the generation of 
```kotlin
private fun messageSize(): Int {
    // ...
}
```
uses the extension function
```kotlin
public inline fun <T> Iterable<T>.sumBy(selector: (T) -> Int): Int {
    // ...
}
```
which is marked as deprecated since Kotlin 1.5. A drop-in replacement for this function is provided (`@ReplaceWith`) and it's called `sumOf()` (which exists since Kotlin 1.4). I had a look at the source code and it seems that these two functions really are the same. The only three differences I could spot were
1. The name (obviously)
2. A superfluous `.toInt()` in the _newer_ (!) `sumOf()`
3. The newer `sumOf()` uses the experimental overload resolution by lambda return type (`@OverloadResolutionByLambdaReturnType`), which means that now there aren't `sumOfDouble()`, `sumOfLong()` and so on, but multiple overloads that only differ in the return type of the passed lambda.

Considering all of this, migrating from `sumBy()` to `sumOf()` should be as trivial as just replacing the name at each usage.
I think it would be a good idea to do this migration soon, since Kotlin might remove the old `sumBy()` soon (or change its semantics), and also all the warnings that are generated are a bit annoying.

Version Data:

- Kotlin/JVM 1.5.21
- OpenJDK 16 (target / language level 11)
- ProtoKt 0.6.2
- ProtoBuf 3.15.3
- gRPC 1.1.0

andrewparmet:
Fixes #117.

### DIFF
--- a/buildSrc/src/main/kotlin/com/toasttab/protokt/gradle/ProtobufBuild.kt
+++ b/buildSrc/src/main/kotlin/com/toasttab/protokt/gradle/ProtobufBuild.kt
@@ -27,10 +27,11 @@ import java.io.File
 import java.net.URLEncoder
 import org.gradle.api.Project
 import org.gradle.api.file.FileCollection
-import org.gradle.api.plugins.JavaPluginConvention
+import org.gradle.api.tasks.SourceSetContainer
 import org.gradle.api.tasks.bundling.Jar
 import org.gradle.internal.os.OperatingSystem
 import org.gradle.kotlin.dsl.apply
+import org.gradle.kotlin.dsl.get
 import org.gradle.kotlin.dsl.named
 import org.gradle.kotlin.dsl.the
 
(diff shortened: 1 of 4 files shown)


buildSrc/src/main/kotlin/com/toasttab/protokt/gradle/ProtobufBuild.kt
```diff
@@ -91,8 +92,9 @@ private fun extraClasspath(project: Project, task: GenerateProtoTask): String {
 private fun configureSources(project: Project, generatedSourcesPath: String) {
     val protoktDir = "$generatedSourcesPath/main/protokt"
 
-    project.the<JavaPluginConvention>()
-        .sourceSets.main.kotlin.srcDirs
+    project.the<SourceSetContainer>()["main"]
+        .java
```
andrewparmet, line 96: I'm not sure if the Java source set is good enough here or if it really does have to be the Kotlin source set. I can't figure out a non-deprecated way of accessing the Kotlin source set.

ogolberg (approved):

PR #123, opened 2021-08-18
Allow parameterized map value wrappers

andrewparmet:
Allows this sort of thing:

```protobuf
message StringListWrapper {
  repeated string value = 1;
}

message AnyListWrapper {
  repeated google.protobuf.Any value = 1;
}

message Foo {
  map<string, StringListWrapper> map_string_list_wrapper = 9 [
    (.protokt.property).value_wrap = "java.util.List<String>"
  ];

  map<string, AnyListWrapper> map_any_list_wrapper = 10 [
    (.protokt.property).value_wrap = "java.util.List<com.toasttab.protokt.Any>"
  ];
}
```

The problem is that we can't arbitrarily nest the parameterizations since the converter has to use raw types:

```kotlin
@AutoService(Converter::class)
object StringListWrapperConverter : Converter<List<String>, StringListWrapper> {
    override val wrapped = StringListWrapper::class

    @Suppress("UNCHECKED_CAST")
    override val wrapper = List::class as KClass<List<String>>

    override fun unwrap(wrapped: List<String>) =
        StringListWrapper { value = wrapped }

    override fun wrap(unwrapped: StringListWrapper) =
        unwrapped.value
}
```

So if you try to do:

```protobuf
message StringListListWrapper {
  repeated StringListWrapper value = 1;
}

message Foo {
  map<string, StringListListWrapper> map_string_list_list_wrapper = 11 [
    (.protokt.property).value_wrap = "java.util.List<List<String>>"
  ];
}
```

you get code that doesn't compile:

```kotlin
StringListWrapperConverter.wrap(it.value) // `it` is a List<List<String>>, not a List<String>
```

So maybe we don't implement this.

I guess we could change the converter interface to something that understands type parameters, but I also have concerns about the cost of copying these lists (or other parameterized types) at serialize/deserialize time since we're always making defensive copies.


andrewparmet:
PoC, not interested in merging.

PR #131, opened 2021-09-21
Bump common protos and exclude protobuf-java to avoid duplicate class issues

andrewparmet:


third-party/proto-google-common-protos/build.gradle.kts
```diff
@@ -20,5 +20,7 @@ pureKotlin()
 enablePublishing()
 
 dependencies {
-    protobuf(libraries.protoGoogleCommonProtos)
+    protobuf(libraries.protoGoogleCommonProtos) {
+        exclude(group = "com.google.protobuf", module = "protobuf-java")
```
ogolberg, line 24: maybe pull the exclude into Dependencies.kt

ogolberg (approved):

third-party/proto-google-common-protos/build.gradle.kts
andrewparmet: Maybe, but one day when we move to version catalogs it'll have to come back.

Issue #154, opened 2022-01-31
Regression from 0.7.2 to 0.7.3 with reserved keywords in protos


pranjalvachaspati-toast:
Something in 0.7.3 makes a proto like this fail to compile:

```
syntax = "proto3";

package foo.bar.baz;

message myMessage {
  oneof my_oneof {
     int32 something = 1;
     int32 null = 2;
  }
}
```

The error is in the class, which should look like the output of 0.7.2:

```
    sealed class MyOneof {
        data class Something(
            val something: Int
        ) : MyOneof()

        data class Null(
            val `null`: Int
        ) : MyOneof()
    }
```

but in 0.7.3 and later looks like this:

```
    sealed class MyOneof {
        data class Something(
            val something: Int
        ) : MyOneof()

        data class Null(
            val ``null``: Int
        ) : MyOneof()
    }
```

PR #155, opened 2022-02-01
Fix oneof with reserved keywords as names

pranjalvachaspati-toast:
This fixes issue #154

### DIFF
--- a/protokt-codegen/src/main/kotlin/com/toasttab/protokt/codegen/annotators/OneofAnnotator.kt
+++ b/protokt-codegen/src/main/kotlin/com/toasttab/protokt/codegen/annotators/OneofAnnotator.kt
@@ -77,13 +77,13 @@ private constructor(
                                 }
                             }
                             .addProperty(
-                                PropertySpec.builder(v.fieldName, v.type)
-                                    .initializer(v.fieldName)
+                                PropertySpec.builder(v.fieldName.removePrefix("`").removeSuffix("`"), v.type)
+                                    .initializer(v.fieldName.removePrefix("`").removeSuffix("`"))
                                     .build()
                             )
                             .primaryConstructor(
                                 FunSpec.constructorBuilder()
-                                    .addParameter(v.fieldName, v.type)
+                                    .addParameter(v.fieldName.removePrefix("`").removeSuffix("`"), v.type)
                                     .build()
                             )
                             .handleSuperInterface(options, v)
--- a/testing/runtime-tests/src/main/proto/toasttab/protokt/testing/rt/oneof/reserved_name.proto
+++ b/testing/runtime-tests/src/main/proto/toasttab/protokt/testing/rt/oneof/reserved_name.proto
@@ -18,8 +18,8 @@ syntax = "proto3";
 package toasttab.protokt.testing.rt.oneof;
 
 message Example {
-    oneof my_oneof {
-        string null = 1;
-        string foo = 2;
-    }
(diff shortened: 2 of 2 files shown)


protokt-codegen/src/main/kotlin/com/toasttab/protokt/codegen/annotators/OneofAnnotator.kt
```diff
@@ -77,13 +77,13 @@ private constructor(
                                 }
                             }
                             .addProperty(
-                                PropertySpec.builder(v.fieldName, v.type)
-                                    .initializer(v.fieldName)
+                                PropertySpec.builder(v.fieldName.removePrefix("`").removeSuffix("`"), v.type)
```
andrewparmet, line 80: I think this is done somewhere else too - is it worth extracting this into a utility?

Separately - I'm not sure why we have to do this. Do you know why things get escaped twice?

testing/runtime-tests/src/main/proto/toasttab/protokt/testing/rt/oneof/reserved_name.proto
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+syntax = "proto3";
+
+package toasttab.protokt.testing.rt.oneof;
+
+message Example {
+    oneof my_oneof {
```
andrewparmet, line 21: nit: two space indents in proto.

protokt-codegen/src/main/kotlin/com/toasttab/protokt/codegen/annotators/OneofAnnotator.kt
pranjalv123: I think it's a kotlinpoet bug - I'm going to look into it, and hopefully then we can get rid of this workaround.

pranjalvachaspati-toast: I think it's a kotlinpoet bug - I'm going to look into it, and hopefully then we can get rid of this workaround.

testing/runtime-tests/src/main/proto/toasttab/protokt/testing/rt/oneof/reserved_name.proto
pranjalvachaspati-toast: fixed

pranjalvachaspati-toast:
Actually it looks like there's a fix for this in a newer version of kotlinpoet already. I'll make a new PR for that.

protokt-codegen/src/main/kotlin/com/toasttab/protokt/codegen/annotators/OneofAnnotator.kt
andrewparmet: Other instances:

https://github.com/open-toast/protokt/blob/main/protokt-codegen/src/main/kotlin/com/toasttab/protokt/codegen/annotators/MessageAnnotator.kt#L102

https://github.com/open-toast/protokt/blob/main/protokt-codegen/src/main/kotlin/com/toasttab/protokt/codegen/annotators/MessageAnnotator.kt#L128

https://github.com/open-toast/protokt/blob/main/protokt-codegen/src/main/kotlin/com/toasttab/protokt/codegen/annotators/DslAnnotator.kt#L68

PR #164, opened 2022-04-14
re-enable API compatibility checks

andrewparmet:


andrewparmet:
This won't pass until I release a new version which I'm going to do after the validation PR passes.

PR #168, opened 2022-05-04
Make runtime multiplatform compatible

andrewparmet:
To do:

- [x] Make protokt-core-lite and protokt-core multiplatform. Requires figuring out how to separate the protobuf-gradle-plugin from Kotlin JVM/Java so that we can invoke local protokt on protokt-core-lite and protokt-core while using Kotlin multiplatform. Looks like pb-and-k [simply ditched the protobuf-gradle-plugin](https://github.com/streem/pbandk/blob/master/buildSrc/src/main/kotlin/ProtocTask.kt). All sorts of problems come up right now, e.g. not resolving the `api` configuration.
    - ~I got this working with [this fork of the protobuf-gradle-plugin](https://github.com/andrewparmet/protobuf-gradle-plugin/tree/support-kotlin-multiplatform). Definitely a hack.~
    - There's an [open issue](https://github.com/google/protobuf-gradle-plugin/issues/497) for supporting Kotlin multiplatform but it looks like it's not in active development.
    - I got this to work with the vanilla plugin by reflectively adding the multiplatform and JS plugins to the list of prerequisite plugins and doing the rest of the configuration locally using public APIs.
- [x] Rework publication configuration to account for multiplatform projects.
    - Using https://github.com/vanniktech/gradle-maven-publish-plugin (which okio and okhttp use).
    - Will have to set new env vars for maven publication: `ORG_GRADLE_PROJECT_mavenCentralUsername` and `ORG_GRADLE_PROJECT_mavenCentralPassword`. Can't get the Gradle properties version to work.
- [x] Fix Android compatibility enforcement for multiplatform projects.
- [x] Which extensions projects can be made multiplatform?
    - Just the API.
- [x] Make proto-google-common-protos multiplatform (the non-gRPC parts).
- [x] (Optional): protokt-gradle-plugin complains about language level 1.4 even though we don't specify it anywhere.
- [x] AutoService can't be used for the well-known types converters
- [x] JS types can't implement function interfaces. Adding a capitalized function with the same name as the class works. ~except that mysteriously the compiler sometimes doesn't resolve calls to the function and instead thinks we're calling the private constructor. https://youtrack.jetbrains.com/issue/KT-52492/Inconsistent-overload-resolution-between-function-and-constructo~ fixed
    - ~Used lower case letters for now. Even with lower case functions the compiler sometimes can't resolve them or throws NoSuchMethodError - makes me think there's a problem in the Gradle config or something having to do with top level functions. Capital letter functions could possibly work with a strategy like this:~
```kotlin
@KtGeneratedMessage("google.protobuf.Duration")
class Duration(dsl: DurationDsl.() -> Unit) : AbstractKtMessage() {
    val seconds: Long
    val nanos: Int
    val unknownFields: UnknownFieldSet

    init {
        val built = DurationDsl().apply(dsl)
        seconds = built.seconds
        nanos = built.nanos
        unknownFields = built.unknownFields
    }
}
```
Exposing a constructor like this probably breaks easy Jackson integration, but we should probably discourage that anyways.
- [x] Get rid of Tag and Numbers classes.
- [x] Get a JS implementation working (conformance tests as well).
- [x] Get protobuf-gradle-plugin working with Kotlin JS plugin and add a JS integration test project.
- [x] Extensions configurations are not quite right - figure out why `protokt.proto` is not propagating through extract-include-protos properly.
- [ ] Remove support for the Kotlin JS plugin and find another way to determine generation of gRPC code for JS projects. Perhaps multiple plugin invocations - once per source set.
- [ ] Upgrade protobuf past 3.24

Unrelated proposed changes for 1.0:
- [x] Bump Kotlin to 1.8.21 minimum (MPP plugin compatibility for IR compiler seems to vary across Kotlin versions).
- [x] ~Change extension wrapper API to work on Sequence<Byte> instead of cloning ByteArray.~ Use read-only views of Bytes directly.
- [x] Default package name convention changes. (Consult grpc-kotlin team for possible fixes to integration with their generated code: https://github.com/grpc/grpc-kotlin/issues/389)


.circleci/config.yml
```diff
@@ -43,10 +46,26 @@ jobs:
...
+
+      - run:
+          name: Install Firefox
+          command: |
+            apk add firefox
 
       - run:
           name: Build and test
-          command: ./gradlew clean check publishToIntegrationRepository --stacktrace
+          command: ./gradlew clean check publishToIntegrationRepository --stacktrace --refresh-dependencies
```
ogolberg, line 68: what does refresh-dependencies do?

buildSrc/build.gradle.kts
```diff
@@ -13,25 +13,89 @@
...
+tasks.withType<KotlinCompile> {
+    dependsOn("generateVersions")
+}
+
+val versionOutputDir = layout.buildDirectory.file("generated/protobuf-version")
+
+sourceSets["main"].java.srcDir(versionOutputDir)
+
+tasks.register("generateVersions") {
+    doFirst {
```
ogolberg, line 48: make these into proper tasks with inputs/outputs declared

alternatively, use the build config plugin (see https://github.com/open-toast/can-i-use-anvil/blob/main/can-i-use-anvil-plugin/build.gradle.kts#L10 for an example)

buildSrc/src/main/kotlin/JavaBasedProjectConventions.kt
```diff
@@ -30,28 +30,24 @@ fun Project.javaBasedProjectConventions() {
...
     }
 
     tasks.withType<KotlinCompile> {
         kotlinOptions {
             allWarningsAsErrors = true
-            jvmTarget = "1.8"
-            freeCompilerArgs = listOf("-Xinline-classes")
-            languageVersion = "1.5"
-            apiVersion = "1.5"
+            freeCompilerArgs = listOf("-Xjvm-default=all")
```
ogolberg, line 42: what does jvm-default do again?

buildSrc/src/main/kotlin/JavaBasedProjectConventions.kt
```diff
@@ -30,28 +30,24 @@ fun Project.javaBasedProjectConventions() {
     }
 
     dependencies {
-        "api"(kotlin("stdlib", "1.5.32"))
```
ogolberg, line 33: this whole file is a good candidate to become a convention plugin; then you can use the regular `api(...)` syntax and other dsls

buildSrc/src/main/kotlin/protokt/v1/gradle/ProtoktBuild.kt
```diff
@@ -0,0 +1,240 @@
...
+                    dependsOn(genProtoTask)
+                }
+            }
+        }
+    }
+
+    linkGenerateProtoTasksAndIncludeGeneratedSource(mainSourceSet, false)
+    linkGenerateProtoTasksAndIncludeGeneratedSource(testSourceSet, true)
+
+    afterEvaluate {
```
ogolberg, line 192: why afterEvaluate?

.circleci/config.yml
andrewparmet: I must have been having some problem getting stale dependencies. I can probably get rid of it now.

buildSrc/src/main/kotlin/JavaBasedProjectConventions.kt
andrewparmet: Tells the Kotlin compiler to generate default methods for non-abstract Kotlin interface members. https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-jvm-default/

This was needed for ABI compatibility for drop-in compatibility before I changed the package naming scheme of the runtime and can probably be removed.

buildSrc/src/main/kotlin/DependencyUtils.kt
```diff
@@ -0,0 +1,60 @@
...
+import org.gradle.api.Project
+import org.gradle.api.artifacts.MinimalExternalModuleDependency
+import org.gradle.api.artifacts.VersionCatalogsExtension
+import org.gradle.api.provider.Provider
+import org.gradle.kotlin.dsl.configure
+import org.gradle.kotlin.dsl.dependencies
+import org.gradle.kotlin.dsl.exclude
+import org.gradle.kotlin.dsl.getByType
+
+fun Project.findLibrary(name: String): String =
```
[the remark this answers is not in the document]
andrewparmet: No dice but I welcome your attempt.

buildSrc/src/main/kotlin/JavaBasedProjectConventions.kt
andrewparmet: I'm not sure I can use the regular syntax since I won't have a Kotlin plugin applied in this file to generate those accessors - this is used both in Android and Kotlin JVM projects.

buildSrc/src/main/kotlin/protokt/v1/gradle/ProtoktBuild.kt
andrewparmet: I probably needed it long ago but it looks like I can get rid of it now.

buildSrc/src/main/kotlin/protokt/v1/gradle/ProtoktBuild.kt
```diff
@@ -0,0 +1,240 @@
...
+        }
+
+    return if (protoktVersion == null) {
+        dependencies.project(":$artifactId")
+    } else {
+        dependencies.create("com.toasttab.protokt:$artifactId:$protoktVersion")
+    }
+}
+
+internal fun Project.isMultiplatform() =
```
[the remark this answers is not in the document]
andrewparmet: I can fix these. I'm going to have to rework the `appliedKotlinPlugin` check since the JS plugin is deprecated and that's how I figure out which gRPC code to generate.

buildSrc/build.gradle.kts
andrewparmet: Done.

buildSrc/src/main/kotlin/DependencyUtils.kt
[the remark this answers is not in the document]
andrewparmet: Nevermind - working now.

buildSrc/src/main/kotlin/JavaBasedProjectConventions.kt
andrewparmet: https://github.com/open-toast/protokt/pull/168/commits/c90f80423981a28ec3cbae8ea55c57a689c35483

buildSrc/src/main/kotlin/JavaBasedProjectConventions.kt
ogolberg: ah ok

ogolberg (approved):

PR #169, opened 2022-05-04
Add prefix to classpath of generated protobuf files 

jeromesklar-toast:
Adds a protokt specific prefix to classpath of generated protobuf files to avoid conflicts in projects that use both protokt and another protobuf library, such as that of google. This will allow projects to safely use protokt and another protobuf library and not have conflicts (which can lead to RuntimeExceptions) between generated file paths.

Note: this is a breaking change and upgrading to this version of protokt will involve making changes to add the "protokt." to certain imports

Testing: Verified this fixes at least one runtime exception that was occurring with google protobuf and protokt libraries in the same service

Future changes:
In the future, we plan on potentially doing some more package cleanup to:
- remove some extraneous "com." prefixes
- remove the special case that converts com.google.protobuf --> com.toasttab.protokt and just universally add the "protokt." prefix solution in all cases. (so, com.google.protobuf --> protokt.com.google.protobuf)


protokt-codegen/src/main/kotlin/com/toasttab/protokt/codegen/impl/PackageResolution.kt
```diff
@@ -115,8 +124,25 @@ private fun javaPackage(respectJavaPackage: Boolean, fileOptions: FileOptions) =
...
-    PPackage.fromString(overrideGoogleProtobuf(type, "com.$rootGoogleProto"))
+// Adds a protokt-specific package prefix to prevent conflicts for projects that use both protokt
+// and another protobuf code generator (e.g. Google, Wire)
+private fun addProtoktPackagePrefix(packageName: String) =
+    "protokt.$packageName"
+
+// Returns the overriden type value or None if no override occurred
+private fun overrideComGoogleProtobuf(type: String): Option<String> {
+    // TODO: make double prefix check not needed by removing com. in advance
+    var overridenType = overrideGoogleProtobuf(type, "com.$rootGoogleProto")
```
andrewparmet, line 135: Typo: `overridden`

testing/android-test-configurations/build.gradle.kts
```diff
@@ -1,3 +1,5 @@
+import com.toasttab.protokt.gradle.protoktExtensions
```
andrewparmet, line 1: Import should go below the license here.

testing/android-test-configurations/build.gradle.kts
```diff
@@ -24,4 +26,6 @@ pureKotlin()
 
 dependencies {
     testImplementation(libraries.protobufLite)
+    testImplementation(project(":testing:android"))
+    protoktExtensions(project(":extensions:protokt-extensions"))
```
andrewparmet, line 30: Are these additions needed? I think we may only need the extensions dependency.

testing/options/src/main/proto/com/toasttab/protokt/testing/options/custom_file_descriptor_object_name.proto
```diff
@@ -15,10 +15,12 @@
 
 syntax = "proto3";
 
-package com.toasttab.protokt.testing.options;
+package toasttab.protokt.testing.options;
```
andrewparmet, line 18: All of these files should be relocated to a directory not containing `com`, so in this case `src/main/proto/toasttab/protokt/...`.

andrewparmet (commented):
Looks great.

testing/android-test-configurations/build.gradle.kts
jeromesklar-toast: build fails without both dependencies

testing/options/src/main/proto/com/toasttab/protokt/testing/options/custom_file_descriptor_object_name.proto
jeromesklar-toast: discussed offline and i'm gonna switch back to com to make diff more readable, and then follow up with moving files/directories around PR

mayabarrett-toast (approved):

testing/android-test-configurations/build.gradle.kts
```diff
@@ -22,6 +25,13 @@ android {
 localProtokt()
 pureKotlin()
 
+protokt {
+    lite = true
+}
+
 dependencies {
+    protoktExtensions(project(":extensions:protokt-extensions-lite"))
```
andrewparmet, line 33: This should be `testProtoktExtensions`.

testing/plugin-options/lite/build.gradle.kts
```diff
@@ -24,7 +24,9 @@ protokt {
 }
 
 dependencies {
+    protoktExtensions(project(":extensions:protokt-extensions-lite"))
```
andrewparmet, line 27: I think this line can be removed.

andrewparmet (commented):
Two tiny things and then this is ready to merge.

buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -42,7 +42,7 @@ object versions {
     const val protoGoogleCommonProtos = "2.8.2"
 
     // Android
-    const val androidGradle = "7.1.0"
+    const val androidGradle = "4.1.0"
```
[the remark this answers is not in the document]
andrewparmet: IntelliJ has a regression that doesn't let you import the project in 2022.1 when it uses AGP 7.1.0. We don't really care what version we're using so we can downgrade until IntelliJ is fixed.

andrewparmet (approved):

ogolberg (changes requested):
Let's punt till 1.0.

* The cost of propagating this backwards-incompatible change through systems that use protokt is substantial.
* Current behavior is consistent with Wire and ScalaPB (other protobuf codegens that target JVM languages)

andrewparmet:
superseded by #168

PR #182, opened 2023-04-27
Add schema descriptor to service descriptor

pranjalvachaspati-toast:


pranjalvachaspati-toast:
We need this to support proto reflection.

andrewparmet:
I see how this is analogous to the way grpc-java populates the schema descriptor field. Right now protokt structures its generated reflection support differently from protobuf-java in a way which conflicts with this PR's approach.

Protokt's generated code composes with itself - code generated with `lite = true` can freely intermix with code generated with `onlyGenerateDescriptors = true`. The same holds for `generateGrpc` and `onlyGenerateGrpc`.

Any schema descriptor registered with the gRPC service descriptors should preserve that property: the gRPC code must compile with `lite = true` and yet have access to the schema descriptor when it's provided on the classpath. An explicit call to the descriptor reference breaks this property. I'm not sure if it's possible to preserve without using Java/Kotlin reflection, but that might be acceptable with some `by lazy` caching.

The broken use case is:
- Project A generates a service with `lite = true` and `generateGrpc = true`
- Project B generates code with `onlyGenerateDescriptors = true`

In this case Project B will not have access to the schema descriptor. It would have to regenerate all protobuf code from Project A with `lite = false` and `generateGrpc = true` and force Project A to modularize differently if it needs to depend on any other code from Project A.

PR #194, opened 2023-09-03
Add integration test for wrapper types

andrewparmet:
Fixes problem where `protoktExtensions` dependencies no longer create a task dependency within a project to create a JAR needed by subsequent protobuf generation.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -123,6 +123,9 @@ workflows:
             parameters:
               jdk: ["8", "11", "17"]
               kotlin: ["1.5.32", "1.6.21", "1.7.20", "1.8.0"]
+            exclude:
+              - jdk: "17"
+                kotlin: "1.5.32"
       - release:
           context: OSS
           filters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -122,10 +122,7 @@ workflows:
           matrix:
             parameters:
               jdk: ["8", "11", "17"]
-              kotlin: ["1.5.32", "1.6.21", "1.7.20", "1.8.0"]
-            exclude:
-              - jdk: "17"
-                kotlin: "1.5.32"
+              kotlin: ["1.6.21", "1.7.21", "1.8.22"]
       - release:
           context: OSS
           filters:
--- a/buildSrc/src/main/kotlin/com/toasttab/protokt/gradle/ProtobufBuild.kt
+++ b/buildSrc/src/main/kotlin/com/toasttab/protokt/gradle/ProtobufBuild.kt
@@ -79,9 +79,11 @@ private fun extraClasspath(project: Project, task: GenerateProtoTask): String {
     // won't pick them up as dependencies unless we do this. There may be a better way to do this but for now just
     // manually do what protobuf-gradle-plugin used to do.
     // https://github.com/google/protobuf-gradle-plugin/commit/0521fe707ccedee7a0b4ce0fb88409eefb04e59d
-    project.tasks.withType<ProtobufExtract>()
-        .filter { it.name.startsWith("extractInclude") }
-        .forEach { it.inputFiles.from(extensions) }
+    project.tasks.withType<ProtobufExtract> {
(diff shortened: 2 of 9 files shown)


gradle-plugin-integration-test/build.gradle.kts
```diff
@@ -30,7 +30,7 @@ buildscript {
 
     dependencies {
         classpath("com.toasttab.protokt:protokt-gradle-plugin:$version")
-        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:${System.getProperty("kotlin.version", "1.4.32")}")
+        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:${System.getProperty("kotlin.version", "1.5.32")}")
```
andrewparmet, line 33: Forgot to change these a while ago and now they're causing problems.

buildSrc/src/main/kotlin/com/toasttab/protokt/gradle/ProtobufBuild.kt
```diff
@@ -73,6 +75,14 @@ private fun extraClasspath(project: Project, task: GenerateProtoTask): String {
         extensions += project.configurations.getByName(TEST_EXTENSIONS)
     }
 
+    // Must explicitly register input files here; if any extensions dependencies are project dependencies then Gradle
+    // won't pick them up as dependencies unless we do this. There may be a better way to do this but for now just
+    // manually do what protobuf-gradle-plugin used to do.
+    // https://github.com/google/protobuf-gradle-plugin/commit/0521fe707ccedee7a0b4ce0fb88409eefb04e59d
+    project.tasks.withType<ProtobufExtract>()
+        .filter { it.name.startsWith("extractInclude") }
+        .forEach { it.inputFiles.from(extensions) }
```
andrewparmet, line 84: I don't love this but not sure of a better way to ensure the extensions files, which are (probably?) declared internally to be output files of their respective jar tasks, are inputs of the various extractInclude tasks used by the protobuf-gradle-plugin internally.

Any ideas?

.circleci/config.yml
```diff
@@ -123,6 +123,9 @@ workflows:
             parameters:
               jdk: ["8", "11", "17"]
               kotlin: ["1.5.32", "1.6.21", "1.7.20", "1.8.0"]
+            exclude:
```
andrewparmet, line 126: Some weird interplay here I didn't want to deal with.

ogolberg: we should probably drop support for kotlin 1.5 at this point

buildSrc/src/main/kotlin/com/toasttab/protokt/gradle/ProtobufBuild.kt
```diff
@@ -73,6 +75,14 @@ private fun extraClasspath(project: Project, task: GenerateProtoTask): String {
         extensions += project.configurations.getByName(TEST_EXTENSIONS)
     }
 
+    // Must explicitly register input files here; if any extensions dependencies are project dependencies then Gradle
+    // won't pick them up as dependencies unless we do this. There may be a better way to do this but for now just
+    // manually do what protobuf-gradle-plugin used to do.
+    // https://github.com/google/protobuf-gradle-plugin/commit/0521fe707ccedee7a0b4ce0fb88409eefb04e59d
+    project.tasks.withType<ProtobufExtract>()
+        .filter { it.name.startsWith("extractInclude") }
```
ogolberg, line 83: hm does this configure the task eagerly? should this be `tasks.withType<> { if (name.startsWith) ... }` instead?

ogolberg (approved):

buildSrc/src/main/kotlin/com/toasttab/protokt/gradle/ProtobufBuild.kt
andrewparmet: Done

.circleci/config.yml
andrewparmet: dropped

PR #204, opened 2023-10-10
platform bytes class

andrewparmet:


protokt-runtime/src/commonMain/kotlin/protokt/v1/AbstractBytes.kt
```diff
@@ -0,0 +1,54 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package protokt.v1
+
+abstract class AbstractBytes internal constructor(
```
andrewparmet, line 18: Seems to be the only way to share this code.

protokt-runtime/src/commonMain/kotlin/protokt/v1/AbstractBytes.kt
```diff
@@ -0,0 +1,54 @@
...
+    final override fun equals(other: Any?) =
+        other is Bytes && value.contentEquals(other.value)
+
+    final override fun hashCode() =
+        value.contentHashCode()
+
+    final override fun toString() =
+        value.contentToString()
+
+    internal companion object {
```
andrewparmet, line 42: Used to hold common code in implementation companions rather than exposing a new public class. The same could be done with AbstractBytes but then the implementation would have to call into this object rather than inheriting the correct behavior. This pattern is already used in AbstractKtMessage and AbstractKtDeserializer, so it feel OK to be consistent with those.

PR #205, opened 2023-10-21
Distinguish between converters that are safe to call on default values and those that would fail during deserialization on default-valued fields

andrewparmet:
Prevents addition of fields that are marked as wrapped from breaking compiled code forwards compatibility when the wrapper type is not safe to use on an absent field. The three out-of-the-box converters that have this property are those for UUID, LocalDate (from String), and InetAddress.

Protokt now has a property that Converters use to indicate their ability to deserialize protobuf default values. Protokt attempts to instantiate each wrapper type with the default value if the wrapped field is not nullable and not marked with the nonnull annotation; i.e., non-optional scalar fields like `bytes`, `string`, `int32`. If that call fails, protokt will insist that the Converter declares itself unable to deserialize the default type and then marks the field as nullable and so it does not attempt to wrap the "zero" instance if the field doesn't appear on the wire.

The non_null option is now able to be applied to non-optional scalar fields in the case where they are wrapped by a converter that does not work on default values. It's a dangerous thing to do, but it's sufficiently hard to do this by accident and I could imagine a real use case for it.

This PR also inverts the type arguments for Converter since I think it's confusing for the Kotlin type to be first with the Protobuf type second when I think of it more as "wrapping from Protobuf type to Kotlin type." Open to undoing this though - it's easy to flip back and forth. Maybe renaming the type parameters is sufficient.


extensions/protokt-jvm-extensions-simple/src/main/kotlin/protokt/v1/LocalDateStringConverter.kt
```diff
@@ -19,28 +19,14 @@ import com.google.auto.service.AutoService
...
     override fun wrap(unwrapped: String): LocalDate =
         LocalDate.parse(unwrapped)
 
     override fun unwrap(wrapped: LocalDate) =
         wrapped.toString()
 }
-
-@AutoService(Converter::class)
-@Deprecated("use LocalDateStringConverter or upgrade protokt")
-object LocalDateConverter : Converter<LocalDate, String> {
```
andrewparmet, line 36: Not needed in V1; only needed for ABI backwards compatibility.

protokt-codegen/src/main/kotlin/protokt/v1/codegen/generate/Wrapper.kt
```diff
@@ -21,90 +21,99 @@ import com.squareup.kotlinpoet.TypeName
...
-            val converter = converter(wrapper, wrapped, ctx)
-            if (converter.optimizedSizeof) {
+        f.withWrapper(ctx) {
+            if (it.optimizedSizeof) {
                 accessSize
             } else {
                 interceptValueAccess(f, ctx, accessSize)
             }
-        } ?: interceptValueAccess(f, ctx, accessSize)
+        } ?: accessSize
```
andrewparmet, line 80: this wasn't doing anything

protokt-codegen/src/main/kotlin/protokt/v1/codegen/generate/Wrapper.kt
```diff
@@ -21,90 +21,99 @@ import com.squareup.kotlinpoet.TypeName
...
-                CodeBlock.of("%T.sizeOf(%L)", converter.className, accessSize)
+        f.withWrapper(ctx) {
+            if (it.optimizedSizeof) {
+                callConverterMethod(OptimizedSizeOfConverter<Any, Any>::sizeOf, it, fieldAccess)
             } else {
-                f.sizeOf(accessSize)
+                accessSize
             }
-        } ?: f.sizeOf(accessSize)
+        } ?: accessSize
```
andrewparmet, line 94: tidy up this call to encapsulate `sizeOf` in MessageSizeGenerator

ogolberg (approved):

andrewparmet:
Prevents addition of fields that are marked as wrapped from breaking compiled code forwards compatibility when the wrapper type is not safe to use on an absent field. The three out-of-the-box converters that have this property are those for UUID, LocalDate (from String), and InetAddress.

Protokt now has a property that Converters use to indicate their ability to deserialize protobuf default values. Protokt attempts to instantiate each wrapper type with the default value if the wrapped field is not nullable and not marked with the nonnull annotation; i.e., non-optional scalar fields like bytes, string, int32. If that call fails, protokt will insist that the Converter declares itself unable to deserialize the default type and then marks the field as nullable and so it does not attempt to wrap the "zero" instance if the field doesn't appear on the wire.

The non_null option is now able to be applied to non-optional scalar fields in the case where they are wrapped by a converter that does not work on default values. It's a dangerous thing to do, but it's sufficiently hard to do this by accident and I could imagine a real use case for it.

This PR also inverts the type arguments for Converter since I think it's confusing for the Kotlin type to be first with the Protobuf type second when I think of it more as "wrapping from Protobuf type to Kotlin type." Open to undoing this though - it's easy to flip back and forth. Maybe renaming the type parameters is sufficient.

PR #212, opened 2023-12-30
implement a converter to dynamic message

andrewparmet:


protokt-reflect/src/jvmMain/kotlin/protokt/v1/google/protobuf/KtMessages.kt
```diff
@@ -0,0 +1,317 @@
...
+        Type.UINT64, Type.FIXED64 -> 0uL
+        Type.UINT32, Type.FIXED32 -> 0u
+        Type.BYTES -> Bytes.empty()
+        else -> field.defaultValue
+    }
+
+fun KtMessage.getField(field: FieldDescriptor) =
+    ProtoktReflect.getField(this, field)
+
+class RuntimeContext internal constructor(
```
andrewparmet, line 74: New public APIs. Open to revision before 1.0.

shared-src/reflect/protokt/v1/reflect/FieldType.kt
```diff
@@ -13,21 +13,15 @@
...
  */
 
-package protokt.v1.codegen.util
+package protokt.v1.reflect
 
 import com.google.protobuf.DescriptorProtos.FieldDescriptorProto.Type
-import com.squareup.kotlinpoet.CodeBlock
-import com.squareup.kotlinpoet.MemberName
-import com.squareup.kotlinpoet.MemberName.Companion.member
-import com.squareup.kotlinpoet.asTypeName
```
andrewparmet, line 22: removing kotlinpoet deps from code now used in reflection at runtime

protokt-codegen/src/main/kotlin/protokt/v1/codegen/generate/DeserializerGenerator.kt
```diff
@@ -31,21 +31,21 @@ import protokt.v1.UnknownFieldSet
...
-import protokt.v1.codegen.util.FieldType.SInt32
-import protokt.v1.codegen.util.FieldType.SInt64
-import protokt.v1.codegen.util.FieldType.UInt32
-import protokt.v1.codegen.util.FieldType.UInt64
 import protokt.v1.codegen.util.KotlinPlugin
 import protokt.v1.codegen.util.Message
 import protokt.v1.codegen.util.Oneof
 import protokt.v1.codegen.util.StandardField
 import protokt.v1.codegen.util.Tag
+import protokt.v1.reflect.FieldType
```
andrewparmet, line 39: All of these types that are used for codegen but are also going to be used in dynamic conversion are moved to shared source and stripped of any bad runtime deps, e.g. KotlinPoet. They're staying `internal` so the public API does not expose them.

protokt-codegen/src/test/kotlin/protokt/v1/codegen/ProtoktCodegenTest.kt
```diff
@@ -30,3 +36,30 @@ class ProtoktCodegenTest : AbstractProtoktCodegenTest() {
             }
     }
 }
+
+object UuidBytesConverter : AbstractConverter<Bytes, UUID>(), OptimizedSizeOfConverter<Bytes, UUID> {
```
andrewparmet, line 40: Allow testing of wrapper types with a debugger.

protokt-reflect/build.gradle.kts
```diff
@@ -0,0 +1,56 @@
...
+
+tasks.withType<JavaCompile> { enabled = true }
+
+sourceSets {
+    main {
+        java {
+            srcDir(rootProject.file("shared-src/reflect"))
+        }
+        proto {
+            srcDir("../extensions/protokt-extensions-lite/src/extensions-proto")
```
andrewparmet, line 49: Generate the protokt extensions with Java for this module to be able to load its descriptor to get our custom options in the reflect runtime.

protokt-reflect/src/jvmMain/kotlin/protokt/v1/google/protobuf/KtMessages.kt
```diff
@@ -0,0 +1,317 @@
...
+
+private fun defaultValue(field: FieldDescriptor) =
+    when (field.type) {
+        Type.UINT64, Type.FIXED64 -> 0uL
+        Type.UINT32, Type.FIXED32 -> 0u
+        Type.BYTES -> Bytes.empty()
+        else -> field.defaultValue
+    }
+
+fun KtMessage.getField(field: FieldDescriptor) =
```
andrewparmet, line 71: `hasField` and `getField` are the two APIs needed to run the Buf validator.

shared-src/reflect/protokt/v1/reflect/ClassLookup.kt
```diff
@@ -55,10 +52,12 @@ class ClassLookup(classpath: List<String>) {
                             .map { classLoader.loadClass(it).kotlin.objectInstance as Converter<*, *> }
                             .toList()
                     }
-            }.run {
-                val table = HashBasedTable.create<String, String, MutableList<Converter<*, *>>>()
-                forEach { table.getOrPut(it.wrapped.qualifiedName!!, it.wrapper.qualifiedName!!) { mutableListOf() }.add(it) }
-                ImmutableTable.builder<String, String, List<Converter<*, *>>>().putAll(table).build()
```
andrewparmet, line 61: Removing Guava from the reflect runtime's dependencies.

shared-src/reflect/protokt/v1/reflect/ClassNames.kt
```diff
@@ -0,0 +1,38 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package protokt.v1.reflect
+
+internal fun inferClassName(className: String, pkg: String): Pair<String, List<String>> {
```
andrewparmet, line 18: These are more or less copied from KotlinPoet since we need them for reflection.

shared-src/reflect/protokt/v1/reflect/FieldType.kt
```diff
@@ -37,28 +31,28 @@ sealed class FieldType {
...
-    data object Fixed64 : Scalar(ULong::class)
-    data object Int32 : Scalar(Int::class)
-    data object Int64 : Scalar(Long::class)
-    data object SFixed32 : Scalar(Int::class)
-    data object SFixed64 : Scalar(Long::class)
-    data object SInt32 : Scalar(Int::class)
-    data object SInt64 : Scalar(Long::class)
-    data object UInt32 : Scalar(UInt::class)
-    data object UInt64 : Scalar(ULong::class)
+    object Bool : Scalar(Boolean::class)
```
andrewparmet, line 43: need compatibility with Kotlin 1.8

third-party/proto-google-common-protos-lite/build.gradle.kts
```diff
@@ -52,3 +52,11 @@ sourceSets {
         }
     }
 }
+
+sourceSets {
+    test {
+        java {
+            srcDir(rootProject.file("shared-src/lite-util"))
```
andrewparmet, line 59: forgotten edge case - good to have the lite tests here.

protokt-reflect/src/jvmMain/kotlin/protokt/v1/google/protobuf/KtMessages.kt
```diff
@@ -0,0 +1,317 @@
...
+            RuntimeContext(getDescriptors(), DEFAULT_CLASS_LOOKUP)
+        }
+
+        fun getContextReflectively() =
+            reflectiveContext
+    }
+}
+
+private fun getDescriptors() =
+    ClassGraph()
```
ogolberg, line 126: can we generate an annotation on the message class that points to the descriptor?

I feel really uncomfortable scanning the whole classpath for this

protokt-reflect/src/jvmMain/kotlin/protokt/v1/google/protobuf/KtMessages.kt
```diff
@@ -0,0 +1,317 @@
...
+
+class RuntimeContext internal constructor(
+    descriptors: Iterable<Descriptors.Descriptor>,
+    private val classLookup: ClassLookup
+) {
+    constructor(descriptors: Iterable<Descriptors.Descriptor>) : this(descriptors, DEFAULT_CLASS_LOOKUP)
+
+    internal val descriptorsByTypeName = descriptors.associateBy { it.fullName }
+
+    fun convertValue(value: Any?) =
```
ogolberg, line 82: not obvious what this public API method does

ogolberg (commented):
See comments inline.

My other question is - can we just codegenerate a converter from KtMesage to DynamicMessage for each KtMessage without using any reflection?

protokt-reflect/src/jvmMain/kotlin/protokt/v1/google/protobuf/KtMessages.kt
andrewparmet: Perhaps. We can't point to it directly, since it won't exist in lite code, but maybe there's a way to make reflection direct instead of scanned.

andrewparmet:
> can we just codegenerate a converter from KtMesage to DynamicMessage for each KtMessage without using any reflection?

We could, but that feels a bit messy. I haven't benchmarked this yet but I think it's a big enough step forward in terms of permitting message validation that it's worth having.

This doesn't prevent us from doing that in the future, of course.

andrewparmet:
Code generation of a converter gets tricky - is there a way to do it without exposing a dependency on protobuf-java?

protokt-reflect/src/jvmMain/kotlin/protokt/v1/google/protobuf/KtMessages.kt
andrewparmet: I can make this not take a nullable argument for starters.

Would naming it better be a good idea? I think it has to be able to take any type since it's called at runtime with whatever field value the descriptor extracts from a protokt message.

A more descriptive name could be `convertProtoktToProtobufJava`.

protokt-reflect/src/jvmMain/kotlin/protokt/v1/google/protobuf/KtMessages.kt
andrewparmet: Okay - the reason this is implemented reflectively is for convenience. If we had an annotation on a message in your hypothetical case, you still have to get those annotations, either by scanning or by explicit reference. If you have them explicitly then you may as well reference the descriptors themselves instead to build your own RuntimeContext.

If we don't want to encourage this kind of scan, which should only be done once, e.g. at application initialization to build validators, then we can remove this code and require users to write this themselves.

protokt-reflect/src/jvmMain/kotlin/protokt/v1/google/protobuf/KtMessages.kt
andrewparmet: The reason this has to be public is to enable [sub-message conversion in protovalidate](https://github.com/andrewparmet/protovalidate-protokt/blob/protokt-v2/src/main/java/build/buf/protovalidate/internal/evaluator/ProtoktImpl.kt). If it's acceptable to convert the whole message every time, even if some portion of the message may never be validated, then this can be internal.

ogolberg (approved):

PR #213, opened 2024-01-04
fix tags in codegen for large field numbers

andrewparmet:
They need to be unsigned in order to be literals.

### DIFF
--- a/protokt-runtime/api/protokt-runtime.api
+++ b/protokt-runtime/api/protokt-runtime.api
@@ -595,7 +595,7 @@ public abstract interface class protokt/v1/KtMessageDeserializer {
 	public abstract fun readSInt32 ()I
 	public abstract fun readSInt64 ()J
 	public abstract fun readString ()Ljava/lang/String;
-	public abstract fun readTag ()I
+	public abstract fun readTag-pVg5ArA ()I
 	public fun readUInt32-pVg5ArA ()I
 	public abstract fun readUInt64-s-VKNKU ()J
 	public abstract fun readUnknown ()Lprotokt/v1/UnknownField;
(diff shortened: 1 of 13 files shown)


testing/interop/src/main/proto/google/protobuf/unittest_proto3.proto
```diff
@@ -0,0 +1,206 @@
...
+}
+
+// TestEmptyMessage is used to test behavior of unknown fields.
+message TestEmptyMessage {}
+
+// TestMessageWithDummy is also used to test behavior of unknown fields.
+message TestMessageWithDummy {
+  // This field is only here for triggering copy-on-write; it's not intended to
+  // be serialized.
+  bool dummy = 536870911;
```
andrewparmet, line 190: This field breaks generated code. The rest of this file will be used to test the implementation of conversion to DynamicMessage.

protokt-runtime/api/protokt-runtime.api
```diff
@@ -595,7 +595,7 @@ public abstract interface class protokt/v1/KtMessageDeserializer {
 	public abstract fun readSInt32 ()I
 	public abstract fun readSInt64 ()J
 	public abstract fun readString ()Ljava/lang/String;
-	public abstract fun readTag ()I
+	public abstract fun readTag-pVg5ArA ()I
```
ogolberg, line 598: what's this?

andrewparmet: UInt nonsense.

ogolberg: is this ABI going to be stable?

aparmet-toast: I don't know. We already do this below - can we merge and revisit pre-1.0?

andrewparmet: I don't know. We already do this below - can we merge and revisit pre-1.0?

ogolberg (approved):

PR #217, opened 2024-01-06
Download prebuilt conformance runner artifacts

ogolberg:
POC for using statically compiled conformance runner binaries from https://github.com/ogolberg/build-protobuf-conformance-runner/releases


testing/conformance/runner/src/test/kotlin/protokt/v1/conformance/ConformanceTest.kt
```diff
@@ -66,7 +66,7 @@ class ConformanceTest {
     fun `run conformance tests`(runner: ConformanceRunner) {
         try {
             command(runner)
-                .runCommand(projectRoot.toPath(), libPathOverride)
```
andrewparmet, line 69: Does this extension function still need a second arg?

andrewparmet (approved):

PR #219, opened 2024-01-14
Add JSON module and enable JSON serialization conformance tests

andrewparmet:


friscoMad:
Hi @andrewparmet, first of all thank you for all the hard work in this lib and sorry to ping you here but my company is considering using protoKT for our stack and I would like to know if you feel that JSON and `protovalidate` integration are planned to be merged "soon", both features + #278 would be very valuable for us.
If we end up deciding going with ProtoKT I have no problem spending some time and helping with the implementation.

andrewparmet:
@friscoMad a few notes:

- The JSON module introduced in this PR is a formality wrapping behavior you can implement yourself if you need it more urgently. Simply make a copy of https://github.com/open-toast/protokt/pull/219/files#diff-b99e1237f1eae191c6e42040e31a8d5ea2bf24d2ae3e2c9bbf2d53a28d59e94d for yourself.
- Protovalidate integration is more complex but I anticipate it being merged sooner - but note that you can copy that for yourself as well.
- I'll address #278 there.

friscoMad:
Thanks for the responses @andrewparmet 
* Wow I did not check the JSON implementation other than seeing a lot of commits and a dozen file changes, but it is great that it is a one-liner.
* I was assuming that for protovalidate we need first to get https://github.com/bufbuild/protovalidate-java/pull/132 this merged but I understand now that that is only required to get some perf updates.

PR #220, opened 2024-03-10
Replace AnimalSniffer with Expediter for Android compat checks

ogolberg:
`extensions-lite` has some `java.time` converters which are not compatible with Android API < 26; `extensions` depends on `extensions-lite` and fails validation


andrewparmet:
Do we still need gummybears at all?

ogolberg:
we don't - expediter just pulls in gummy bears internally

andrewparmet (approved):

PR #246, opened 2024-04-10
Fix an edge case in enum prefix stripping

ogolberg:
If, after stripping the common prefix, the enum name starts with a digit - i.e. it's not a valid Protobuf name - the kotlin code, if escaped, will compile to java but fail to compile to javascript

the solution is to only strip the common prefix, if all stripped enum names remain valid protobuf names (which is equivalent to them not starting with a digit)


andrewparmet:
What about stripping the prefix and prefixing the enum name with an underscore? I don't remember if we still do any underscore escaping of property names...

ogolberg:
I don't have a preference either way. The logic is a bit simpler with the current change; as in, we either strip or we do not strip the prefix.

PR #251, opened 2024-04-22
rename core classes

andrewparmet:


protokt-core/src/jvmMain/kotlin/protokt/v1/google/protobuf/Anies.kt
```diff
@@ -13,34 +13,34 @@
  * limitations under the License.
  */
 
-@file:JvmName("AnyUtil")
+@file:JvmName("Anies")
```
andrewparmet, line 16: https://github.com/protocolbuffers/protobuf/blob/main/java/kotlin/src/main/kotlin/com/google/protobuf/Anies.kt

🤮

ogolberg (approved):

PR #271, opened 2024-05-28
Remove support for the Kotlin JS plugin

### LINKED ISSUE #208: Kotlin JS plugin is deprecated
Devise a scheme to generate common/JVM/JS code, perhaps via multiple protobuf plugin invocations - once per platform, declaring source sets for that platform.

andrewparmet:
Reworks Kotlin JS and multiplatform support to run the protoc plugin one time per target, placing generated code in a separate directory per target.

Fixes #208.


buildSrc/src/main/kotlin/LocalProtoktBuild.kt
```diff
@@ -23,9 +23,7 @@ import protokt.v1.gradle.configureProtokt
 import java.io.File
 
 fun Project.localProtokt(disableJava: Boolean = true) {
-    configureProtokt(this, null, disableJava) {
-        "$rootDir/protokt-codegen/build/install/$CODEGEN_NAME/bin/$CODEGEN_NAME"
-    }
+    configureProtokt(this, null, disableJava, "$rootDir/protokt-codegen/build/install/$CODEGEN_NAME/bin/$CODEGEN_NAME")
```
andrewparmet, line 26: No need for deferred evaluation anymore and this was causing a problem in the plugin integration test as it tried to create the new `_codegen_` configuration once per target.

extensions/protokt-extensions-lite/build.gradle.kts
```diff
@@ -42,7 +40,12 @@ protokt {
 
 kotlin {
     sourceSets {
-        val commonMain by getting {}
+        // todo: not sure why this is needed
```
andrewparmet, line 43: I thought I preserved the behavior where we add the core dependencies automatically - and it works for JVM code - but attempting to publish multiplatform `common` code runs into problems unless I add this manually. I think we can deal with this later.

gradle-plugin-integration-test/multiplatform/build.gradle.kts
```diff
@@ -73,6 +72,7 @@ kotlin {
         all {
             compilations.all {
                 kotlinOptions {
+                    allWarningsAsErrors = false
```
andrewparmet, line 75: The suppression of `OPTIONAL_DECLARATION_USAGE_IN_NON_COMMON_SOURCE` _itself_ throws a compilation warning that can't (as far as I can tell) be excluded with a compiler arg. Gets a shrug from me.

gradle-plugin-integration-test/multiplatform/build.gradle.kts
```diff
@@ -87,13 +87,6 @@ tasks.named<Test>("jvmTest") {
     useJUnitPlatform()
 }
 
-// awkward that we have to apply the plugin after source sets are configured
```
andrewparmet, line 90: See code comment about determining Kotlin targets without `afterEvaluate`.

protokt-codegen/src/main/kotlin/protokt/v1/codegen/Main.kt
```diff
@@ -63,14 +63,12 @@ private fun main(`in`: InputStream, out: OutputStream) {
 
     val grpcKotlinFiles = generateGrpcKotlinStubs(params, req)
 
-    if (files.isNotEmpty() || grpcKotlinFiles.isNotEmpty()) {
```
andrewparmet, line 66: If you generate nothing and the input has a file that uses proto3 `optional`, protoc fails since you didn't reply that you support the feature. We just didn't happen to trigger this. IIRC I added this check to handle an issue we had when trying to return no files, but clearly that's not a problem anymore.

protokt-codegen/src/main/kotlin/protokt/v1/codegen/generate/FileGenerator.kt
```diff
@@ -33,8 +34,19 @@ private class FileGenerator(
                 contents.info.kotlinPackage,
                 fileName(contents)
             ).apply {
-                addDeprecationSuppression()
-                // https://github.com/square/kotlinpoet/pull/533
+                addAnnotation(
+                    AnnotationSpec.builder(Suppress::class).apply {
+                        addMember("DEPRECATION".embed())
+                        // Suppresses a failure due to usage of @JvmStatic in common code. It seems prudent to add this
```
andrewparmet, line 40: This only actually triggered a problem in the multiplatform integration test project. So strange.

protokt-codegen/src/main/kotlin/protokt/v1/codegen/generate/ServiceGenerator.kt
```diff
@@ -431,15 +430,14 @@ private class ServiceGenerator(
         }
 
     private fun <T> pivotPlugin(jvm: T, js: T) =
```
andrewparmet, line 432: Now that we have the concept of a generation target I want to clean this up, but it's complex logic and doesn't have to be cleaned up yet.

shared-src/gradle-plugin/protokt/v1/gradle/ProtoktBuild.kt
```diff
@@ -128,8 +160,16 @@ private fun Project.linkGenerateProtoToSourceCompileForKotlinMpp(mainSourceSetNa
...
         from(fileTree("${layout.buildDirectory.get()}/extracted-protos/main"))
-        duplicatesStrategy = DuplicatesStrategy.EXCLUDE // TODO: figure out how to get rid of this
+        excludeDuplicates()
     }
+
+    val jsProcessResources = tasks.findByName("jsProcessResources") as Copy?
+    jsProcessResources?.excludeDuplicates()
+}
+
+// TODO: figure out how to get rid of this?
```
andrewparmet, line 170: Can't quite figure out what's going wrong here.

shared-src/gradle-plugin/protokt/v1/gradle/ProtoktBuild.kt
```diff
@@ -139,8 +179,11 @@ private fun Project.linkGenerateProtoTasksAndIncludeGeneratedSource(sourceSetNam
 
     extension.generateProtoTasks.ofSourceSet(protoSourceSetRoot).forEach { genProtoTask ->
         configureSourceSets(sourceSetName, protoSourceSetRoot, genProtoTask)
+
+        // todo: it would be better to avoid this by making the outputs of genProtoTask an input of the correct compile task
```
andrewparmet, line 183: Would also like to do this correctly if at all possible.

extensions/protokt-extensions-lite/build.gradle.kts
ogolberg: okay, expand the todo comment

extensions/protokt-extensions/build.gradle.kts
```diff
@@ -45,6 +43,9 @@ kotlin {
         val commonMain by getting {
             dependencies {
                 api(project(":extensions:protokt-extensions-lite"))
+
+                // todo: not sure why this is needed
```
ogolberg, line 47: ditto

ogolberg (approved):

extensions/protokt-extensions-lite/build.gradle.kts
andrewparmet: Figured this out.

PR #352, opened 2025-02-26
Final updates to runtime and codegen fixes

andrewparmet:
Uncovered a few bugs when starting to demonstrate some internal upgrades.
- A few message interface implementation bugs
- Removed files break Buf breaking checks

Puts back a non-null accessor generator option for oneof fields.

New extension number reservation has been approved: https://github.com/protocolbuffers/protobuf/pull/20485

### DIFF
--- a/buildSrc/src/main/kotlin/RemotePublishing.kt
+++ b/buildSrc/src/main/kotlin/RemotePublishing.kt
@@ -89,7 +89,7 @@ fun Project.enablePublishing(defaultJars: Boolean = true) {
                 configure(KotlinMultiplatform(JavadocJar.Empty()))
             }
             pluginManager.withPlugin(KotlinPlugins.JVM) {
-                configure(KotlinJvm(JavadocJar.Empty(), true))
+                configure(KotlinJvm(JavadocJar.Empty()))
             }
         }
 
--- a/buildSrc/src/main/kotlin/protokt.third-party-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/protokt.third-party-conventions.gradle.kts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 Toast, Inc.
+ * Copyright (c) 2022 Toast, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -13,6 +13,4 @@
  * limitations under the License.
  */
 
-subprojects {
-    group = "${rootProject.group}.thirdparty"
-}
+group = "com.toasttab.protokt.thirdparty"
--- a/protokt-codegen/src/main/kotlin/protokt/v1/codegen/generate/Implements.kt
+++ b/protokt-codegen/src/main/kotlin/protokt/v1/codegen/generate/Implements.kt
@@ -27,24 +27,15 @@ import protokt.v1.codegen.generate.Nullability.nonNullPropName
 import protokt.v1.codegen.util.Message
 import protokt.v1.codegen.util.StandardField
 import kotlin.reflect.KClass
-import kotlin.reflect.KProperty
 
 internal object Implements {
-    class OverriddenProperty(
-        val property: KProperty<*>
-    )
-
     fun StandardField.overrides(
         ctx: Context,
         msg: Message
     ) =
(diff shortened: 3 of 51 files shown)


buildSrc/src/main/kotlin/RemotePublishing.kt
```diff
@@ -89,7 +89,7 @@ fun Project.enablePublishing(defaultJars: Boolean = true) {
                 configure(KotlinMultiplatform(JavadocJar.Empty()))
             }
             pluginManager.withPlugin(KotlinPlugins.JVM) {
-                configure(KotlinJvm(JavadocJar.Empty(), true))
+                configure(KotlinJvm(JavadocJar.Empty()))
```
andrewparmet, line 92: This is the default value.

buildSrc/src/main/kotlin/protokt.third-party-conventions.gradle.kts
```diff
@@ -13,6 +13,4 @@
  * limitations under the License.
  */
 
-subprojects {
-    group = "${rootProject.group}.thirdparty"
-}
+group = "com.toasttab.protokt.thirdparty"
```
andrewparmet, line 16: This was getting clobbered by the common-conventions plugin - we've been publishing to the wrong group since that change went in.

protokt-codegen/src/main/kotlin/protokt/v1/codegen/generate/Implements.kt
```diff
@@ -61,7 +70,7 @@ internal object Implements {
         }
 
     private fun TypeSpec.Builder.delegateProperties(msg: Message, ctx: Context, canonicalName: String, fieldName: String) {
-        val fieldsByName = msg.fields.filterIsInstance<StandardField>().associateBy { it.fieldName }
```
andrewparmet, line 64: Don't exclude overrides that are oneofs.

protokt-codegen/src/main/kotlin/protokt/v1/codegen/generate/Implements.kt
```diff
@@ -77,7 +77,7 @@ internal object Implements {
                     field.name,
                     (field.returnType.classifier as KClass<*>).asTypeName()
                         .let {
-                            if (assumeNotNull) {
+                            if (assumeNotNull && !field.returnType.isMarkedNullable) {
```
andrewparmet, line 80: Can't delegate a field and mark is as non-nullable if the delegated field is nullable in the interface.

protokt-codegen/src/main/kotlin/protokt/v1/codegen/generate/MessageGenerator.kt
```diff
@@ -159,6 +159,9 @@ private class MessageGenerator(
                             .build()
                     )
                 }
+                if (property.overrides) {
```
andrewparmet, line 162: When the name of a field is the same as the field on the interface and we generate a non-null accessor for it, the override needs to be on the generated delegating property, not the non-null accessor.

protokt-codegen/src/main/kotlin/protokt/v1/codegen/util/GeneratorContext.kt
```diff
@@ -56,7 +56,6 @@ private fun generateFdpObjectNames(files: List<FileDescriptorProto>): Map<String
         Pair(
             fdp.name,
             fdp.fileOptions.protokt.fileDescriptorObjectName.takeIf { it.isNotEmpty() }
-                ?: fdp.fileOptions.default.javaOuterClassname.takeIf { it.isNotEmpty() }
```
andrewparmet, line 59: Ignore all Java options.

protokt-codegen/src/main/kotlin/protokt/v1/codegen/generate/Nullability.kt
```diff
@@ -43,11 +45,7 @@ internal object Nullability {
                 type !in setOf(FieldType.Message, FieldType.Enum)
 
     fun propertyType(o: Oneof) =
-        if (o.generateNonNullAccessor) {
```
andrewparmet, line 46: This was previously non-functional.

ogolberg (approved):

PR #398, opened 2025-06-23
Adopt some stuff from Google

andrewparmet:
- Fully define unmodifiable collections to allow no-copy creation of message builders from existing messages
- Introduce annotation that must be opted into to help prevent unintended usage of the generated message runtime


protokt-codegen/src/main/kotlin/protokt/v1/codegen/generate/FileGenerator.kt
```diff
@@ -47,6 +49,12 @@ private class FileGenerator(
                     }.build()
                 )
 
+                addAnnotation(
+                    AnnotationSpec.builder(ClassName.bestGuess("kotlin.OptIn"))
+                        .addMember("$PROTOKT_V1.OnlyForUseByGeneratedProtoCode::class")
```
andrewparmet, line 54: The compiler doesn't actually let you reference these by class literal. Weird language design decision.

ogolberg (approved):

PR #492, opened 2026-03-17
Fix LazyConvertingMap hashCode instability breaking AndroidX DataStore

andrewparmet:
The anonymous Map.Entry objects in ConvertingEntrySet didn't override hashCode()/equals(), so AbstractMap.hashCode() used identity-based hashes that changed on every call. This caused AndroidX DataStore to throw "Data in DataStore was mutated but DataStore is only compatible with Immutable types" on the second write.

Add proper hashCode/equals/toString to the Map.Entry implementation following the Map.Entry contract. Tighten LazyConvertingMap and BuilderScope map type parameters to non-nullable bounds.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,3 +2,4 @@ org.gradle.jvmargs=-Xmx4096M
 kotlin.mpp.stability.nowarn=true
 org.gradle.caching=true
 android.experimental.enableTestFixturesKotlinSupport=true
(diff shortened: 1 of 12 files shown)


ogolberg (approved):

pfccole2025-svg:
WTF Delbert Cole ur just never going to stop b n sneaky lying

PR #512, opened 2026-04-14
Migrate benchmarks to kotlinx-benchmark for JVM+Native

andrewparmet:
- Migrate from JMH to kotlinx-benchmark, which delegates to JMH on JVM and uses its own runtime on Kotlin/Native
- Add macosArm64 native benchmark target alongside JVM
- Fix benchmark property names: use `protokt.v1.codec` and `protokt.v1.collection.factory` (was missing the `.v1` prefix)


buildSrc/src/main/kotlin/RuntimeFriendPaths.kt
```diff
@@ -33,46 +34,46 @@ import java.io.File
...
+    friendPaths(":protokt-runtime")
+
     configure<KotlinMultiplatformExtension> {
         compilerOptions {
             freeCompilerArgs.add("-opt-in=protokt.v1.OnlyForUseByGeneratedProtoCode")
         }
+    }
+}
 
+fun Project.friendPaths(vararg friendProjectPaths: String) {
```
andrewparmet, line 46: Since native can't use system properties to configure codec/collection factory, we have to use friend paths now. This structure decouples friend path declaration from the runtime module so we can expose the `internal` members necessary to set codec/coll. fac. from the benchmark projects.

ogolberg (approved):

PR #533, opened 2026-05-27
Implement protovalidate laziness via ProtoktMessage wrapper for CEL navigation

andrewparmet:
Alternative to #402.

Wraps protokt messages as `com.google.protobuf.Message` for CEL navigation. `MessageReflector.celValue()` returns a custom `ProtoktMessage` that implements the subset of `MessageOrBuilder` cel-java's `DescriptorMessageProvider` actually calls (`getField`, `hasField`, `getRepeatedFieldCount`, `getRepeatedField`, `hasOneof`, `getOneofFieldDescriptor`, `getDescriptorForType`, `getDefaultInstanceForType`, `getUnknownFields`). Everything else throws `UnsupportedOperationException` since CEL never asks.

Nested message fields return another `ProtoktMessage` so the navigation stays lazy: only the fields the validation rules touch get adapted.

Compared to #402:
- Same lazy-access behavior; eliminates the eager DynamicMessage walk.
- Drops the `StructValue` path entirely.
- Drops the three cel-java patches (https://github.com/andrewparmet/cel-java/tree/fix/fixed32-fixed64-handling, https://github.com/andrewparmet/cel-java/tree/fix/fieldmask-message-preservation, https://github.com/andrewparmet/cel-java/tree/fix/protomessagevalueprovider-hierarchy) as a hard prerequisite.
- Drops the `enableCelValue(true)` requirement on protovalidate-java's side.
- WKT-typed fields still go through `context.convertValue` because cel-java's `ProtoAdapter` casts WKTs to their concrete Java protobuf classes.


ogolberg:
I like the laziness, but do you that implementing an interface partially is too fragile?

andrewparmet:
It is a little fragile. The folks at Buf are looking at a shim interface we could share. We can hold off here until they get back to me.

PR #544, opened 2026-07-12
Carry generateProto task dependency on generated source directories

### LINKED ISSUE #545: Publishing fails gradle implicit task dependency check
I ran into an issue when trying to publish an api module of mine that had protos. This is the error I got.
```
Task ':iosArm64SourcesJar' uses this output of task ':generateProto'
without declaring an explicit or implicit dependency.
```

I would like to avoid having to put explicit task dependencies in my gradle configuration. The library should be able to handle this by itself.

dronda-t:
I was running into this issue when trying to publish an api module of mine that had protos
  Task ':iosArm64SourcesJar' uses this output of task ':generateProto'
  without declaring an explicit or implicit dependency.

Adding the generated output directories to Kotlin source sets as bare Directory providers loses the producer task information, so tasks that consume the source set without an explicit dependsOn (e.g. the KMP <target>SourcesJar tasks) fail Gradle's implicit-dependency validation with the above error.

My fix is to wrap the source directories in a file collection built by the generate task so every consumer of the source set inherits the dependency. i tested this fix with my own gradle project and it solved the issue.

Closes: https://github.com/open-toast/protokt/issues/545

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,10 +106,10 @@ jobs:
         name: integration-repository
         path: build/repos/integration
     - name: Build and test
-      run: ./gradlew "-Dkotlin-integration.version=${{ matrix.kotlin }}" "-Djava-integration.version=${{ matrix.jdk }}" build --stacktrace --no-daemon
+      run: ./gradlew "-Dkotlin-integration.version=${{ matrix.kotlin }}" "-Djava-integration.version=${{ matrix.jdk }}" build publishAllPublicationsToIntegrationTestRepository --stacktrace --no-daemon
       working-directory: gradle-plugin-integration-test
     - name: Verify configuration cache compatibility
-      run: ./gradlew "-Dkotlin-integration.version=${{ matrix.kotlin }}" "-Djava-integration.version=${{ matrix.jdk }}" --configuration-cache build --stacktrace --no-daemon
+      run: ./gradlew "-Dkotlin-integration.version=${{ matrix.kotlin }}" "-Djava-integration.version=${{ matrix.jdk }}" --configuration-cache build publishAllPublicationsToIntegrationTestRepository --stacktrace --no-daemon
       working-directory: gradle-plugin-integration-test
 
   # AGP 9 built-in Kotlin: runs with the default Kotlin version only, since AGP 9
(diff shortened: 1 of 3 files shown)


gradle-plugin-integration-test/multiplatform/build.gradle.kts
```diff
@@ -83,6 +83,12 @@ tasks.named<Test>("jvmTest") {
     useJUnitPlatform()
 }
 
+// Adding this dependnecy helps verify that Gradle's
```
andrewparmet: What is this doing exactly?

shared-src/gradle-plugin/protokt/v1/gradle/ProtoktBuild.kt
```diff
@@ -251,13 +251,15 @@ private fun Project.linkGenerateProtoTasksAndIncludeGeneratedSource(target: Kotl
 
     generateProtoTask?.let { genProtoTask ->
         // Only include this target's output directory, not all targets' output directories.
+        // Attach the generate task via builtBy so that every consumer of the source set
```
andrewparmet: Probably don't need this comment - `builtBy` is self-describing.

gradle-plugin-integration-test/multiplatform/build.gradle.kts
dronda-t: It was a very barebones attempt at testing this. I'm not sure to what extent you want to test this. Please let me know what you think is best. I'm about to push up something a little more involved.

.github/workflows/ci.yml
```diff
@@ -106,10 +106,10 @@ jobs:
         name: integration-repository
         path: build/repos/integration
     - name: Build and test
-      run: ./gradlew "-Dkotlin-integration.version=${{ matrix.kotlin }}" "-Djava-integration.version=${{ matrix.jdk }}" build --stacktrace --no-daemon
+      run: ./gradlew "-Dkotlin-integration.version=${{ matrix.kotlin }}" "-Djava-integration.version=${{ matrix.jdk }}" build publishAllPublicationsToIntegrationTestRepository --stacktrace --no-daemon
```
andrewparmet, line 109: Could we publish to Maven local instead? The integration repository is specifically for the integration jobs - unless we ought to use the artifact in them?

PR #553, opened 2026-08-23
Centralize dependency repositories and formatting

andrewparmet:
- Define dependency repositories once for the build and reject project-level repository declarations.
- Apply project-scoped Spotless configuration through the common conventions plugin so parent projects do not rescan children.


settings.gradle.kts
```diff
@@ -25,6 +25,26 @@ buildscript {
 
 apply(plugin = "net.vivin.gradle-semantic-build-versioning")
 
+dependencyResolutionManagement {
+    repositories {
+        mavenCentral()
+        google()
+        maven("https://packages.jetbrains.team/maven/p/krpc/grpc")
+        ivy {
```
ogolberg: this only applies to the conformance module

andrewparmet: True - would be lovely to see it in Maven Central.

ogolberg (approved):

PR #556, opened 2026-08-23
Unify malformed input failures

andrewparmet:
Add ProtoktDecodeException as the common malformed-input failure.

Use it for invalid tags and wire types, truncation, invalid lengths, overlong varints, recursion limits, unsupported groups, and partially consumed nested messages.


protokt-runtime-kotlinx-io/src/commonMain/kotlin/protokt/v1/KotlinxIoSourceReader.kt
```diff
@@ -139,13 +144,14 @@ internal class KotlinxIoSourceReader(
             shift += 7
         }
         // discard upper bits for oversized varints
-        while (true) {
+        repeat(5) {
```
ogolberg, line 147: is this intentional?

andrewparmet: Yes - it's a good find from upstream: https://github.com/protocolbuffers/protobuf/blob/889dd941e662fa8cfd26b328daf3765b64237456/java/core/src/main/java/com/google/protobuf/CodedInputStream.java#L1182-L1208

which is an unrolled version of this loop - max ten bytes for a varint per the protobuf spec.

ogolberg (approved):

PR #581, opened 2026-08-24
Stabilize integration build inputs

andrewparmet:
## Summary

- Restrict integration-test Kotlin and protobuf formatting to checked-in source directories.
- Prevent format checks from traversing generated build directories while protobuf tasks replace them.
- Use fixed Unix line endings and pin formatted source types to LF on checkout, including Windows.
- Install both the baseline JDK 17 toolchain and the selected matrix JDK so integration builds do not download toolchains at execution time.

## Failures addressed

- Windows integration builds failed `spotlessKotlinGradleCheck` because Git checked out `build.gradle.kts` with CRLF while Spotless required LF. The failure reported every line as changed, starting with `-/*\r\n` and `+/*\n` ([AGP 9 failure](https://github.com/open-toast/protokt/actions/runs/32686475822/job/97315102252), [Kotlin 2.3.21 failure](https://github.com/open-toast/protokt/actions/runs/32686475822/job/97315102655)).
- A subsequent Windows configuration-cache build failed `spotlessKotlinGradle` with `NoSuchFileException` while traversing `android-test-fixtures/build/generated/java/generateReleaseProto/...`; protobuf generation replaced that output while Spotless was snapshotting it ([failure](https://github.com/open-toast/protokt/actions/runs/32687647315/job/97318108193)).
- Integration rows selected JDK 21 or 25 while the build also requested a JDK 17 toolchain, which led to intermittent failure.

## Acceptance criteria

- Integration builds can run protobuf generation and format checks in parallel without a missing generated-directory failure.
- Format checks do not depend on compiler output while determining line endings.
- Checked-in Kotlin, Kotlin Gradle, and protobuf sources pass the same checks on every CI operating system.
- Every integration row has its required Java toolchains before Gradle starts.


gradle-plugin-integration-test/.gitattributes
```diff
@@ -0,0 +1,3 @@
+*.kt text eol=lf
+*.kts text eol=lf
+*.proto text eol=lf
```
ogolberg, line 3: should we just set autocrlf=false?

.github/workflows/ci.yml
```diff
@@ -100,7 +100,9 @@ jobs:
       uses: actions/setup-java@v5
       with:
         distribution: corretto
-        java-version: ${{ matrix.jdk }}
+        java-version: |
+          17
```
ogolberg: do we need to install 17 here, or just in the integration test job?

gradle-plugin-integration-test/.gitattributes
andrewparmet: We could - core.autocrlf is local Git configuration so it would have to be set in jobs, AFAIK.

``` patch
  steps:
   - name: Configure git line endings
     run: git config --global core.autocrlf false
   - uses: actions/checkout@v7
```

.github/workflows/ci.yml
andrewparmet: As structured this job needs both - wrapper-types pins its toolchain to Java 17, and the other integration projects use matrix.jdk.

andrewparmet: Looking at an alternative

ogolberg (approved):

PR #584, opened 2026-08-24
Declare the codegen binary input

andrewparmet:
- Model the installed codegen executable and its producer as a declared Gradle input.
- Generate a platform-specific launcher at a stable build-directory path.
- Remove execution-time plugin-path mutation and implicit task dependency wiring.
- Apply the same producer relationship to the repository’s local codegen build.

## Effects

- Every protobuf generation task depends on the task or configuration that produces its codegen executable.
- Windows and Unix generation invoke a stable launcher without mutating protobuf task options in `doFirst`.


ogolberg (approved):
this is clever

PR #588, opened 2026-09-04
Memoize messageSize with a -1 sentinel instead of a Lazy delegate

shmuelrosansky-toast:
## Summary

Every generated message held `override val messageSize: Int by lazy { messageSize() }`. Per instance that is a `SynchronizedLazyImpl` (24 B), a capturing initializer lambda (16 B, plus one synthetic class per message type) and, once read, a boxed `Integer`, on top of the reference field. The generated class now keeps a single transient `Int` initialized to `-1` and computes the size on first read, the same shape as protobuf-java's `memoizedSize`.

- Codegen only. Generated code relies on nothing new in the runtime, so 0.12.3-generated jars run on a 0.12.2 runtime and vice versa. `KtMessage` is unchanged.
- The unsynchronized write is a benign race: every thread computes the same value and `Int` writes are atomic. `-1` is the sentinel because an empty message legitimately has size 0.
- The field name is added to the codegen's reserved words so a proto field with that name is renamed, as `messageSize` already is.
- `@Transient` keeps field-based reflection (Gson, Jackson field visibility) from emitting the memo.
- Second commit is benchmark-only: enables the JMH GC profiler so `gc.alloc.rate.norm` is reported.

This is the base of a small stack; #589 (exact-size lists) and #590 (right-sized maps) build on it.

## Generated code before / after (`Timestamp`)

```kotlin
-    override val messageSize: Int by lazy { messageSize() }
+    @Transient
+    private var memoizedMessageSize: Int = -1
+
+    override val messageSize: Int
+        get() {
+            var size = memoizedMessageSize
+            if (size == -1) {
+                size = messageSize()
+                memoizedMessageSize = size
+            }
+            return size
+        }
```

## Measurements

Retained heap of the parsed JMH datasets, measured with JOL (`GraphLayout.parseInstance`, JDK 11, exact and deterministic):

| Arm | medium dataset (101 `GenericMessage1`) | small dataset (101 `GenericMessage4`) |
|---|---:|---:|
| 0.12.2 | 3,187,792 B / 114,677 objects | 13,592 B / 510 objects |
| this PR | 2,791,248 B / 98,154 objects (-12.4%) | 8,728 B / 307 objects (-36%) |

JMH `gc.alloc.rate.norm`, deserialize benchmarks: medium 4.51-4.66 MB/op -> 4.27 MB/op, small 24,120 B/op -> 19,272 B/op (-20%). Serialize benchmarks unchanged (sizes are memoized after the first iteration either way). Timings were within error bars on all six benchmarks.

Class count: `protokt-core-lite` 466 -> 409 classes, `proto-google-common-protos-lite` 903 -> 787 (every `$messageSize$2` Lazy initializer class gone).

## Verification

- `./gradlew clean check publishToIntegrationRepository` on JDK 11: green.
- `gradle-plugin-integration-test` with Kotlin 1.8.22 and 2.2.0: green.
- New `testing/runtime-tests/.../MessageSizeTest`: size equals serialized size (populated, empty, with unknown fields, with a map), repeated reads stable, concurrent first reads agree, Jackson output has no memo field.

Motivation on our side: heap dumps of Toast's Android POS showed one `SynchronizedLazyImpl` plus lambda per resident protokt message.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/protokt-codegen/src/main/kotlin/com/toasttab/protokt/codegen/annotators/MessageAnnotator.kt
+++ b/protokt-codegen/src/main/kotlin/com/toasttab/protokt/codegen/annotators/MessageAnnotator.kt
@@ -46,6 +46,8 @@ import com.toasttab.protokt.rt.KtGeneratedMessage
 import com.toasttab.protokt.rt.KtMessage
 import com.toasttab.protokt.rt.UnknownFieldSet
 
+internal const val MEMOIZED_MESSAGE_SIZE = "memoizedMessageSize"
+
 class MessageAnnotator
 private constructor(
     private val msg: Message,
@@ -138,11 +140,34 @@ private constructor(
         handleSuperInterface(msg, ctx)
     }
 
+    // Memoized with a sentinel rather than `by lazy` so each message holds one Int instead of a
+    // SynchronizedLazyImpl, an initializer lambda, and a boxed result. The unsynchronized write is
+    // a benign race: every thread computes the same value, and Int writes are atomic on the JVM.
     private fun TypeSpec.Builder.handleMessageSize() =
         addProperty(
+            PropertySpec.builder(MEMOIZED_MESSAGE_SIZE, Int::class)
+                .addModifiers(KModifier.PRIVATE)
+                .mutable(true)
+                .addAnnotation(Transient::class)
+                .initializer("-1")
+                .build()
+        ).addProperty(
             PropertySpec.builder("messageSize", Int::class)
                 .addModifiers(KModifier.OVERRIDE)
-                .delegate("lazy { messageSize() }")
+                .getter(
+                    FunSpec.getterBuilder()
+                        .addCode(
+                            """
+                                |var size = $MEMOIZED_MESSAGE_SIZE
+                                |if (size == -1) {
+                                |    size = messageSize()
+                                |    $MEMOIZED_MESSAGE_SIZE = size
+                                |}
+                                |return size
+                            """.bindMargin()
+                        )
+                        .build()
+                )
                 .build()
         )
 
--- a/testing/runtime-tests/src/test/kotlin/com/toasttab/protokt/testing/rt/MessageSizeTest.kt
+++ b/testing/runtime-tests/src/test/kotlin/com/toasttab/protokt/testing/rt/MessageSizeTest.kt
@@ -1,0 +1,106 @@
+/*
+ * Copyright (c) 2026 Toast Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.toasttab.protokt.testing.rt
+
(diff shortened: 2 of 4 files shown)


shmuelrosansky-toast:
Per-commit attribution, in case reviewers want to take this piecemeal. Retained size is JOL on the parsed medium dataset (deterministic); alloc/op is the JMH GC profiler on `deserializeMediumFromMemory` (3 forks; the per-fork spread is JIT escape analysis, not the change).

| Arm | JOL retained (medium) | objects | alloc B/op |
|---|---:|---:|---:|
| 0.12.2 | 3,187,792 | 114,677 | 4.51-4.66 MB |
| + commit 1 (sentinel) | 2,791,248 (-12.4%) | 98,154 | 4.27 MB |
| + commit 2 (exact lists) | 2,560,544 (-8.3% more) | 93,908 | 4.39 MB |
| + commit 3 (right-sized maps) | 2,509,080 (-2.0% more) | 93,908 | 4.64-4.80 MB |

Commit 1 lowers both retention and churn. Commit 2 trades one extra exactly-sized array per non-empty repeated field (about +3% churn) for -8% retained. Commit 3 copies every map of up to five entries, which costs +6-9% churn for -2% retained and no object-count change. I'd understand dropping commit 3; the first two stand on their own. Timing on the six benchmarks was within error bars in single-fork runs; the only timing signal I could resolve on a shared machine was the full branch being roughly 7% slower than 0.12.2 on deserializeMedium, and I could not attribute that between commits 2 and 3 with confidence.

shmuelrosansky-toast:
Restructured into a stack: this PR is now the sentinel change only (plus the benchmark profiler commit). The exact-size list change moved to #589 and the right-sized map change to #590, each stacked on the previous one. The attribution table above still applies commit by commit; commit 2 there is #589 and commit 3 is #590.

testing/runtime-tests/src/test/kotlin/com/toasttab/protokt/testing/rt/MessageSizeTest.kt
```diff
@@ -0,0 +1,106 @@
...
+    @Test
+    fun `repeated reads return the same value`() {
+        val first = nested.messageSize
+        repeat(10) {
+            assertThat(nested.messageSize).isEqualTo(first)
+        }
+    }
+
+    @Test
+    fun `concurrent first reads agree`() {
```
andrewparmet, line 81: Not 100% sold on the value of these two tests, but they're probably not harmful.

protokt-codegen/src/main/kotlin/com/toasttab/protokt/codegen/annotators/MessageAnnotator.kt
```diff
@@ -138,11 +140,34 @@ private constructor(
         handleSuperInterface(msg, ctx)
     }
 
+    // Memoized with a sentinel rather than `by lazy` so each message holds one Int instead of a
```
andrewparmet, line 143: Not sure we need this comment.

andrewparmet (commented):
Looks good.

protokt-codegen/src/main/kotlin/com/toasttab/protokt/codegen/annotators/MessageAnnotator.kt
shmuelrosansky-toast: 👍  removed

testing/runtime-tests/src/test/kotlin/com/toasttab/protokt/testing/rt/MessageSizeTest.kt
shmuelrosansky-toast: 🤷‍♂️  removed, there are other tests here that cover this

andrewparmet (approved):

PR #591, opened 2026-09-04
Memoize serializedSize with a -1 sentinel instead of a Lazy delegate

shmuelrosansky-toast:
## Summary

Every generated message held `private val __serializedSize: Int by lazy { ... }`. Per instance that is a `SynchronizedLazyImpl` (24 B), a capturing initializer lambda (16 B, plus one synthetic class per message type) and, once read, a boxed `Integer`, on top of the reference field. The generated class now keeps a single transient `Int` initialized to `-1`; the size computation moves into a private `__computeSerializedSize()` and `serializedSize()` memoizes it on first call, the same shape as protobuf-java's `memoizedSize`.

- Codegen only. Generated code relies on nothing new in the runtime; `Message`, `AbstractMessage` and the codec interfaces are unchanged.
- The unsynchronized write is a benign race: every thread computes the same value and `Int` writes are atomic. `-1` is the sentinel because an empty message legitimately has size 0.
- `@Transient` is `kotlin.jvm.Transient`, which is an optional expectation in common code, so multiplatform targets are unaffected; on the JVM it keeps field-based reflection (Gson, Jackson field visibility) from emitting the memo.
- The checked-in bootstrap types were regenerated with `:protokt-codegen:regenerateBootstrap` (license header kept as before). The bootstrap diff looks large because git aligns the moved size-computation body; the only change is the sentinel shape.

This is the `main` counterpart of open-toast/protokt#588 on the 0.x line.

## Generated code before / after (`protokt.v1.FileOptions` from the bootstrap)

```kotlin
-  private val __serializedSize: Int by lazy {
-    var result = 0
-    if (_fileDescriptorObjectName.wireValue().isNotEmpty()) {
-      result += sizeOf(10u) + sizeOf(_fileDescriptorObjectName.wireValue())
-    }
-    result += unknownFields.size()
-    result
-  }
+  @Transient
+  private var __serializedSize: Int = -1
...
-  override fun serializedSize(): Int =
-    __serializedSize
+  private fun __computeSerializedSize(): Int {
+    var result = 0
+    if (_fileDescriptorObjectName.wireValue().isNotEmpty()) {
+      result += sizeOf(10u) + sizeOf(_fileDescriptorObjectName.wireValue())
+    }
+    result += unknownFields.size()
+    return result
+  }
+
+  override fun serializedSize(): Int {
+    var size = __serializedSize
+    if (size == -1) {
+      size = __computeSerializedSize()
+      __serializedSize = size
+    }
+    return size
+  }
```

## Measurements

Same change measured on the 0.x line (#588) with JOL on the parsed JMH datasets: retained heap of 101 parsed `GenericMessage1` went from 3,187,792 B / 114,677 objects to 2,791,248 B / 98,154 objects (-12.4%), and 101 small messages from 13,592 B / 510 objects to 8,728 B / 307 objects (-36%). Deserialize allocation per op dropped 6-20% depending on message size; serialize was unchanged. The object layout is identical on `main`, so the per-instance saving (40 B and two objects per message, one class per message type) carries over.

Motivation on our side: heap dumps of Toast's Android POS showed ~7,400 resident protokt v1 messages, each holding a `SynchronizedLazyImpl` plus lambda.

## Verification

- `:protokt-codegen:test`, `:protokt-runtime:jvmTest`, `:testing:protokt-generation:test` green on JDK 17; `spotlessApply` clean.
- New `testing/protokt-generation/.../SerializedSizeTest`: size equals serialized byte count (populated, with a map, empty, with unknown fields), repeated reads stable, concurrent first reads agree.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/protokt-codegen/src/main/kotlin/protokt/v1/codegen/generate/MessageGenerator.kt
+++ b/protokt-codegen/src/main/kotlin/protokt/v1/codegen/generate/MessageGenerator.kt
@@ -269,13 +269,29 @@ private class MessageGenerator(
     private fun dereferenceNullableBackingProperty(propName: String, oneof: Boolean) =
         "requireNotNull($propName) { \"$propName is assumed non-null with (protokt.v1.${if (oneof) "oneof" else "property"}).generate_non_null_accessor but was null\" }".bindSpaces()
 
+    // Memoized with a sentinel rather than `by lazy` so each message holds one Int instead of a
+    // SynchronizedLazyImpl, an initializer lambda, and a boxed result. The unsynchronized write is
+    // a benign race: every thread computes the same value, and Int writes are atomic.
     private fun TypeSpec.Builder.handleMessageSize(propertySpecs: List<PropertySpec>, propertyInfoList: List<PropertyInfo>) {
-        addProperty(generateMessageSize(msg, propertySpecs, ctx, propertyInfoList))
+        addProperty(
+            PropertySpec.builder(SERIALIZED_SIZE, Int::class)
+                .addModifiers(KModifier.PRIVATE)
+                .mutable(true)
+                .addAnnotation(Transient::class)
+                .initializer("-1")
+                .build()
+        )
+        addFunction(generateComputeSerializedSize(msg, propertySpecs, ctx, propertyInfoList))
         addFunction(
             buildFunSpec("serializedSize") {
                 returns(Int::class)
                 addModifiers(KModifier.OVERRIDE)
-                addStatement("return $SERIALIZED_SIZE")
+                addStatement("var·size·=·$SERIALIZED_SIZE")
+                beginControlFlow("if·(size·==·-1)")
+                addStatement("size·=·$COMPUTE_SERIALIZED_SIZE()")
+                addStatement("$SERIALIZED_SIZE·=·size")
+                endControlFlow()
+                addStatement("return·size")
             }
         )
     }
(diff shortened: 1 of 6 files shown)


protokt-codegen/src/main/kotlin/protokt/v1/codegen/generate/MessageGenerator.kt
```diff
@@ -269,13 +269,29 @@ private class MessageGenerator(
     private fun dereferenceNullableBackingProperty(propName: String, oneof: Boolean) =
         "requireNotNull($propName) { \"$propName is assumed non-null with (protokt.v1.${if (oneof) "oneof" else "property"}).generate_non_null_accessor but was null\" }".bindSpaces()
 
+    // Memoized with a sentinel rather than `by lazy` so each message holds one Int instead of a
```
andrewparmet, line 272: Same as 0.x PR - not sure we need this comment.

testing/protokt-generation/src/test/kotlin/protokt/v1/testing/SerializedSizeTest.kt
```diff
@@ -0,0 +1,88 @@
...
+    @Test
+    fun `repeated reads return the same value`() {
+        val first = populated.serializedSize()
+        repeat(10) {
+            assertThat(populated.serializedSize()).isEqualTo(first)
+        }
+    }
+
+    @Test
+    fun `concurrent first reads agree`() {
```
andrewparmet, line 73: Likewise, unsure of the value of these tests.

andrewparmet (commented):
Fun to see the impact in the bootstrap code.

andrewparmet (approved):

